PR #6, opened 2016-10-05
Add bulk operations

DzmitryKavalets:


src/main/java/com/epam/ta/reportportal/core/launch/impl/DeleteLaunchHandler.java
```diff
@@ -94,6 +97,19 @@ public OperationCompletionRS deleteLaunch(String launchId, String projectName, S
 		return new OperationCompletionRS("Launch with ID = '" + launchId + "' successfully deleted.");
 	}
 
+	@Override
+	public List<OperationCompletionRS> deleteLaunches(String[] ids, String projectName, String userName) {
+		final List<String> toDelete = asList(ids);
+		final List<Launch> launches = launchRepository.find(toDelete);
```
avarabyeu: you can use stream here to iterate over this launches only once

DzmitryKavalets: Updated

PR #11, opened 2016-10-10
Add jacoco reporting to measure test coverage of external tests

tkachoff:


build.gradle
```diff
@@ -145,6 +150,27 @@ task wrapper(type: Wrapper) {
     gradleVersion = '3.0'
 }
 
+//Start section for external jacoco test report
+tasks.withType(JavaCompile) {
+    if (System.getProperty("DEBUG", "false") == "true") {
```
avarabyeu: Maybe it would be better to read this property once and assign to some global variable

PR #14, opened 2016-10-13
Allow userLogin in recipients field

DzmitryKavalets:


src/main/java/com/epam/ta/reportportal/core/project/impl/UpdateProjectHandler.java
```diff
@@ -99,6 +96,16 @@
 	@Autowired
 	private ApplicationEventPublisher publisher;
 
+	@Autowired
+	public void setProjectRepository(ProjectRepository projectRepository) {
```
avarabyeu: Constructor injection would be better

DzmitryKavalets: Updated

PR #15, opened 2016-10-19
Epmrpp 20932

DzmitryKavalets:


src/main/java/com/epam/ta/reportportal/ws/resolver/PagingHandlerMethodArgumentResolver.java
```diff
@@ -38,6 +38,11 @@
  */
 public class PagingHandlerMethodArgumentResolver extends org.springframework.data.web.PageableHandlerMethodArgumentResolver {
 
+	public static final String ALL_PROJECTS_INFO_METHOD = "getAllProjectsInfo";
```
avarabyeu: do not mix default values with argument resolving. Use @Pageable annotation to specify default values

PR #46, opened 2017-02-28
migrated db to different names in serverSettings

pictureb00k:
added tests for controller of analytics settings

### DIFF
--- a/src/main/resources/migration/v3_0_0_2.js
+++ b/src/main/resources/migration/v3_0_0_2.js
@@ -1,0 +1,10 @@
+db.serverSettings.update(
+    {},
+    {$rename: {"serverEmailConfig":"serverEmailDetails"}})
+db.serverSettings.update(
+    {},
+    {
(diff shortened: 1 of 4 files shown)


src/main/resources/migration/v3_0_0_2.js
```diff
@@ -0,0 +1,10 @@
+db.serverSettings.update(
+    {},
+    {$rename: {"serverEmailConfig":"serverEmailDetails"}})
+db.serverSettings.update(
+    {},
+    {
+        $set: {"_class": "com.epam.ta.reportportal.database.entity.settings.ServerSettings",
+            "serverEmailDetails._class": "com.epam.ta.reportportal.database.entity.settings.ServerEmailDetails"}
+    }
+)
```
avarabyeu: Please, split to two scripts. Make sure field is present before updating it:
db.serverSettings.update(
    {'_class': {$exists : false}}},
    {
        $set: {"_class": "com.epam.ta.reportportal.database.entity.settings.ServerSettings"}
    }
)

PR #50, opened 2017-03-06
Added contributors

pictureb00k:


src/main/java/com/epam/ta/reportportal/info/AnalyticsInfoContributor.java
```diff
@@ -0,0 +1,57 @@
...
+    public AnalyticsInfoContributor(ServerSettingsRepository settingsRepository) {
+        this.settingsRepository = settingsRepository;
+    }
+
+    @Override
+    public Map<String, Object> contribute() {
+        Optional<Map<String, AnalyticsDetails>> analytics = Optional.ofNullable(settingsRepository.findOne("default"))
+                .flatMap(settings -> Optional.ofNullable(settings.getAnalyticsDetails()));
+        Map<String, Object> result = new HashMap<>();
+        analytics.ifPresent(it -> result.put("analytics", it));
```
avarabyeu: map -> convert to map -> 
if not present -> return empty map

PR #57, opened 2017-03-22
update release dockerfile

pictureb00k:


docker/Dockerfile-release
```diff
@@ -1,17 +1,17 @@
-FROM frolvlad/alpine-oraclejdk8:slim
+FROM openjdk:alpine
```
avarabyeu: Please, specify more precise tag here. I assume '8-jre-alpine' is the right one

PR #67, opened 2017-04-16
statistics calculation for test-based strategy

pictureb00k:


src/main/java/com/epam/ta/reportportal/core/statistics/TestBasedStatisticsFacade.java
```diff
@@ -64,4 +67,24 @@ public TestItem deleteExecutionStatistics(TestItem testItem) {
             return testItem;
         }
     }
+
+    @Override
+    public TestItem identifyStatus(TestItem testItem) {
+        if (testItem.getType().sameLevel(TestItemType.TEST)) {
+            List<TestItem> allDescendants = testItemRepository.findAllDescendants(testItem.getId());
```
avarabyeu: why we can search for skipped/failed items directly in db?

PR #147, opened 2017-07-14
Fixed not deleting demo data's attachments by cleaning job

pictureb00k:


src/main/java/com/epam/ta/reportportal/demo_data/TestBasedDemoDataFacade.java
```diff
@@ -77,7 +77,8 @@
         }).collect(toList());
     }
 
-    private List<String> generateStories(Map<String, Map<String, List<String>>> storiesStructure, int i, String launchId) {
+    private List<String> generateStories(Map<String, Map<String, List<String>>> storiesStructure, int i,
```
avarabyeu: what does i mean? let's rename to count

PR #151, opened 2017-07-21
Implemented watch latest unique launches

pictureb00k:


src/main/java/com/epam/ta/reportportal/ws/controller/impl/LaunchController.java
```diff
@@ -229,6 +229,27 @@ public LaunchResource getLaunch(@PathVariable String projectName, @PathVariable
...
+    @ResponseBody
+    @ResponseStatus(OK)
+    @ApiOperation("Get latest unique launches for specified project by filter")
+	public Iterable<LaunchResource> getLatestLaunches(@PathVariable  String projectName, @FilterFor(Launch.class) Filter filter,
+            @SortFor(Launch.class) Pageable pageable) {
+		return getLaunchMessageHandler.getLatestLaunches(normalizeId(projectName), filter, pageable);
+	}
+
+	@Override
+    @RequestMapping(value = "/latest/mode", method = GET)
```
avarabyeu: why mode  is not path variable. Unclear: mode means 'debug'

pictureb00k: they have different access level
in debug mode only project members have control

PR #161, opened 2017-07-31
Latest widgets + Bug fixing + External Systems

pictureb00k:
Contains fixed bug with statistics.
Methods of taking issue types from external systems


src/main/java/com/epam/ta/reportportal/core/widget/content/OverallStatisticsContentLoader.java
```diff
@@ -84,4 +96,34 @@ private String getFieldName(String fieldName) {
...
 		return split[split.length - 1];
 	}
+
+    /**
+     * Excludes unwanted filter conditions for latest launches mode.
+     *
+     * @param filter filter
+     * @return optimized filter
+     */
+    private Filter optimizeFilter(Filter filter) {
```
avarabyeu: let's give some more readable name.

PR #192, opened 2017-09-06
More activities

pictureb00k:


src/main/java/com/epam/ta/reportportal/core/imprt/ImportLaunchHandlerImpl.java
```diff
@@ -25,37 +25,62 @@
...
-        String launch = strategy.importLaunch(projectId, userName, file);
-
+        File tempFile = transferToTempFile(file);
+        eventPublisher.publishEvent(new ImportStartedEvent(projectId, userName, file.getOriginalFilename()));
+        String launch = strategy.importLaunch(projectId, userName, tempFile);
+        eventPublisher.publishEvent(new ImportFinishedEvent(projectId, userName, file.getOriginalFilename()));
         return new OperationCompletionRS("Launch with id = " + launch + " is successfully imported.");
     }
+
+    private File transferToTempFile(MultipartFile file) {
```
avarabyeu: why do we transfer to separate file?

PR #203, opened 2017-09-20
Widget fix

pictureb00k:


src/main/java/com/epam/ta/reportportal/core/widget/content/CasesTrendContentLoader.java
```diff
@@ -157,4 +160,22 @@
...
+	 *
+	 * @param input - callback output result
+	 * @return - transformed Map with reverse ordered elements
+	 */
+	private Map<String, List<ChartObject>> mapRevert(Map<String, List<ChartObject>> input, Sort sorting) {
+		Sort.Order order = sorting.iterator().next();
+		if (order.getProperty().equalsIgnoreCase(SORT_FIELD) && order.isAscending()) {
+			return input;
+		}
+		return input.entrySet().stream().collect(MoreCollectors.toLinkedMap(Map.Entry::getKey, it -> {
```
avarabyeu: does this map immutable or we can just reverse all map items without creating new map?

src/main/java/com/epam/ta/reportportal/core/widget/impl/WidgetUtils.java
```diff
@@ -84,4 +85,16 @@ public static void checkUniqueName(String newWidgetName, List<Widget> existingWi
...
 	}
+
+	/**
+	 * Used by trend charts to reorder results for trend representation.
+	 *
+	 * @param sort Trend Chart sorting
+	 * @return true if need to be reverted
+	 */
+	public static boolean needRevert(Sort sort) {
+		String property = sort.iterator().next().getProperty();
```
avarabyeu: what if iterator has no next value? is it possible?

PR #224, opened 2017-10-19
Auto analyze

pictureb00k:


src/main/java/com/epam/ta/reportportal/core/log/impl/AsyncCreateLogHandler.java
```diff
@@ -71,12 +74,12 @@ public EntryCreatedRS createLog(@Nonnull SaveLogRQ createLogRQ, MultipartFile fi
...
         Log log = logBuilder.get().addSaveLogRQ(createLogRQ).addTestItem(testItem).build();
+		logRepository.save(log);
 
         try {
-            logRepository.save(log);
             logIndexer.indexLog(log);
-        } catch (Exception exc) {
-            throw new ReportPortalException("Error while Log instance creating.", exc);
+        } catch (Exception e) {
+            LOGGER.warn("Log '" + log.getId() + "' is not indexed. Error: \n{} ", e);
```
avarabyeu: use pattern. see slf4j docs

PR #238, opened 2017-10-31
fix not removing tmp import file [TO MASER]

pbortnik:


src/main/java/com/epam/ta/reportportal/core/imprt/impl/junit/XunitImportStrategy.java
```diff
@@ -75,6 +75,10 @@ public String importLaunch(String projectId, String userName, File file) {
             return processZipFile(file, projectId, userName);
         } catch (IOException e) {
             throw new ReportPortalException(ErrorType.BAD_IMPORT_FILE_TYPE, file.getName(), e);
+        } finally {
+            if (null != file) {
+                file.delete();
```
avarabyeu: can we try-catch it and just log error message in case if delete is not successful

PR #246, opened 2017-11-16
Ignore analyzer

pbortnik:


src/main/java/com/epam/ta/reportportal/core/imprt/impl/junit/XunitImportStrategy.java
```diff
@@ -106,6 +110,11 @@ private String processZipFile(File zip, String projectId, String userName) throw
 			finishLaunch(launchId, projectId, userName, processResults(futures));
 			return launchId;
 		} catch (InterruptedException | ExecutionException | TimeoutException | IllegalArgumentException e) {
+			if (savedLaunchId != null) {
+				Launch launch = launchRepository.findOne(savedLaunchId);
+				launch.setStartTime(new Date());
+				launchRepository.partialUpdate(launch);
```
avarabyeu: partial update does not make sense here since 'launch' object has all the fields populated from by .findOne method. 
Please consider the following example that update date only:
```java
Launch launch = new Launch;
launch.setStatistics(null);
//set all non-null fields to null

launch.setDate(Calendar.getInstance().getTime());
repo.partialUpdate(launch);
```

src/main/java/com/epam/ta/reportportal/core/launch/impl/DeleteLaunchHandler.java
```diff
@@ -109,10 +109,10 @@ public OperationCompletionRS deleteLaunches(String[] ids, String projectName, St
 		final User user = userRepository.findOne(userName);
 		final Project project = projectRepository.findOne(projectName);
 		launches.forEach(launch -> validate(launch, user, project));
-		launchRepository.delete(toDelete);
 		launches.forEach(launch -> logIndexer.cleanIndex(projectName,
```
avarabyeu: what if logindex returns error? wouldn't be better to wrap it up with some try/catch?

pbortnik: log index is wrapped with try/catch in the place where it could be error

https://github.com/reportportal/service-api/blob/25e094248b40b6a3a4a68451344b4d6a588633b3/src/main/java/com/epam/ta/reportportal/core/analyzer/client/AnalyzerServiceClient.java#L110

PR #280, opened 2018-01-22
fix widgets update script

pbortnik:


src/main/java/com/epam/ta/reportportal/migration/ChangeSets_4_0.java
```diff
@@ -45,13 +46,19 @@
...
+				where("contentOptions.type").is("trends_chart").and("contentOptions.gadgetType").is("statistic_trend")
+		));
 		query.fields().include("_id");
 		mongoTemplate.stream(query, DBObject.class, collection).forEachRemaining(widget -> {
 			Update update = new Update();
+			String type = (String) ((DBObject) widget.get("contentOptions")).get("type");
 			update.set("contentOptions.type", "trends_chart");
 			update.set("contentOptions.gadgetType", "statistic_trend");
-			update.set("contentOptions.widgetOptions", new BasicDBObject("viewMode", new String[] { "barMode" }));
+			String viewMode = type.equals("line_chart") ? "areaChartMode" : "barMode";
```
avarabyeu: equals on constant!
do we have constants for the modes?

src/main/java/com/epam/ta/reportportal/migration/ChangeSets_4_0.java
```diff
@@ -45,13 +46,19 @@
...
+			update.set("contentOptions.widgetOptions", new BasicDBObject("viewMode", new String[] { viewMode }));
```
avarabyeu: no need to create new array for each iteration

src/main/java/com/epam/ta/reportportal/migration/ChangeSets_4_0.java
pbortnik: no, we have not

PR #293, opened 2018-03-27
custom rules for analyzer

pbortnik:


src/main/java/com/epam/ta/reportportal/core/analyzer/strategy/AnalyzeItemsStrategy.java
```diff
@@ -0,0 +1,36 @@
...
+
+import com.epam.ta.reportportal.database.entity.item.TestItem;
+
+import java.util.List;
+
+/**
+ * @author Pavel Bortnik
+ */
+@FunctionalInterface
+public interface AnalyzeItemsStrategy {
```
avarabyeu: is it really strategy? seems to be like a getter. Can we rename it?

src/main/java/com/epam/ta/reportportal/core/configs/AnalyzeStrategyConfig.java
```diff
@@ -0,0 +1,80 @@
...
+	};
+
+	private AnalyzeItemsStrategy MANUALLY_ANALYZED_STRATEGY = (project, launchId) -> {
+		List<TestItem> itemsByAutoAnalyzedStatus = testItemRepository.findItemsByAutoAnalyzedStatus(false, launchId);
+		logIndexer.cleanIndex(project, itemsByAutoAnalyzedStatus.stream().map(TestItem::getId).collect(Collectors.toList()));
+		return itemsByAutoAnalyzedStatus;
+	};
+
+	@Bean
+	public Map<AnalyzeItemsMode, AnalyzeItemsStrategy> analyzerModeMapping() {
```
avarabyeu: it would be better to explicitly set name of bean with type Map

PR #317, opened 2018-05-16
Ps migrations filestorage

dshybeka:
Added simple implementations of data stores
 - LocalDataStore - store files locally
 - SeaweedDataStore - store files in seaweed fs

### DIFF
--- a/src/main/java/com/epam/ta/reportportal/store/filesystem/LocalFilePathGenerator.java
+++ b/src/main/java/com/epam/ta/reportportal/store/filesystem/LocalFilePathGenerator.java
@@ -1,11 +1,15 @@
 package com.epam.ta.reportportal.store.filesystem;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.stereotype.Component;
 
 import java.util.UUID;
 
 @Component
 public class LocalFilePathGenerator {
+
+	private static final Logger LOG = LoggerFactory.getLogger(LocalFilePathGenerator.class);
 
 	/**
 	 * Generate relative file path for new local file.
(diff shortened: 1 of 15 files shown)


src/main/java/com/epam/ta/reportportal/store/filesystem/LocalFilePathGenerator.java
```diff
@@ -0,0 +1,28 @@
...
+	public String generate() {
+
+		String uuid = UUID.randomUUID().toString();
+
+		String levelOne = uuid.substring(0, 2);
+		String levelTwo = uuid.substring(2, 4);
+		String levelThree = uuid.substring(4, 6);
+		String tail = uuid.substring(6);
+
+		System.out.println(levelOne + levelTwo + levelThree + tail);
```
avarabyeu: logger, please!

src/test/resources/test-application.properties
```diff
@@ -6,3 +6,5 @@ spring.datasource.platform=postgresql
 spring.jpa.properties.hibernate.jdbc.lob.non_contextual_creation=true
 spring.jpa.hibernate.ddl-auto=none
 spring.jpa.hibernate.ddl=false
+
+datastore.default.path=d:\\epam\\report-portal\\data\
```
avarabyeu: can we use some temp dir here?

src/main/java/com/epam/ta/reportportal/store/filesystem/LocalFilePathGenerator.java
dshybeka: sure, will update, thanks

src/test/resources/test-application.properties
dshybeka: sure, this is test properties file, will update property to datastore.default.path=/data/storage as in main applicaiton.properties.

PR #335, opened 2018-06-28
AutoAnalyzed flag fix

Yumfriez:
AutoAnalyzed flag will depend on data, that was received from Client, instead of always being "false". If no data received from Client, flag will be "false" by default.

https://github.com/reportportal/reportportal/issues/427


src/main/java/com/epam/ta/reportportal/core/item/UpdateTestItemHandlerImpl.java
```diff
@@ -125,8 +125,6 @@ public UpdateTestItemHandlerImpl(TestItemRepository testItemRepository, Statisti
 				verifyTestItem(testItem, issueDefinition.getId());
 				TestItem before = SerializationUtils.clone(testItem);
 
-				//if item is updated then it is no longer auto analyzed
-				issueDefinition.getIssue().setAutoAnalyzed(false);
```
pbortnik: it would not work correctly
we decided that if an item has 'true' flag than after any update of the item, the flag should be changed into 'false' condition.
possible solution is to ask ui side to send flag 'false' on any update rq

PR #336, opened 2018-06-29
Fix for personal project index removing in ElasticSearch.

Yumfriez:
Now, after removing user, his personal project index will be removed from ES too.

### DIFF
--- a/src/main/java/com/epam/ta/reportportal/core/user/impl/DeleteUserHandler.java
+++ b/src/main/java/com/epam/ta/reportportal/core/user/impl/DeleteUserHandler.java
@@ -24,12 +24,14 @@ package com.epam.ta.reportportal.core.user.impl;
 import com.epam.ta.reportportal.auth.UatClient;
 import com.epam.ta.reportportal.commons.Predicates;
 import com.epam.ta.reportportal.commons.validation.BusinessRule;
+import com.epam.ta.reportportal.core.analyzer.ILogIndexer;
 import com.epam.ta.reportportal.core.user.IDeleteUserHandler;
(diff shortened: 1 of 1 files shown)


src/main/java/com/epam/ta/reportportal/core/user/impl/DeleteUserHandler.java
```diff
@@ -79,6 +84,8 @@ public OperationCompletionRS deleteUser(String userId, String principal) {
 			throw new ReportPortalException("Error while deleting user", exp);
 		}
 
+		logIndexer.deleteIndex(user.getLogin().toLowerCase()+PersonalProjectService.PERSONAL_PROJECT_POSTFIX.toLowerCase());
```
pbortnik: please, install and use our project formatter from [here](https://github.com/reportportal/reportportal/blob/master/idea_formatting_profile.xml)

src/main/java/com/epam/ta/reportportal/core/user/impl/DeleteUserHandler.java
```diff
@@ -79,6 +84,8 @@ public OperationCompletionRS deleteUser(String userId, String principal) {
 			throw new ReportPortalException("Error while deleting user", exp);
 		}
 
+		logIndexer.deleteIndex(user.getLogin().toLowerCase() + PersonalProjectService.PERSONAL_PROJECT_POSTFIX.toLowerCase());
```
avarabyeu: this is not right way to get personal project name. Use ProjectRepository.findPersonalProjectName

PR #348, opened 2018-07-23
Duration widget

pbortnik:


src/main/resources/application-dev-win.yaml
```diff
@@ -6,7 +6,7 @@ spring:
       host: localhost
 rp:
   mongo:
-    uri: mongodb://localhost:27017
+    uri: mongodb://dev.epm-rpp.projects.epam.com:27017
```
avarabyeu: no dev server names at public

PR #372, opened 2018-09-25
Flaky widget fix

Yumfriez:


src/main/java/com/epam/ta/reportportal/core/widget/content/loader/FlakyCasesTableContentLoader.java
```diff
@@ -31,10 +36,14 @@
 	public Map<String, ?> loadContent(List<String> contentFields, Map<Filter, Sort> filterSortMapping, Map<String, String> widgetOptions,
 			int limit) {
 
+		validateWidgetOptions(widgetOptions);
+
 		validateFilterSortMapping(filterSortMapping);
 
 		Filter filter = GROUP_FILTERS.apply(filterSortMapping.keySet());
 
+		filter.withCondition(new FilterCondition(Condition.EQUALS, false, widgetOptions.get(LAUNCH_NAME_FIELD), NAME));
```
pbortnik: In the result of flaky widget should be info about latest launch. I can't see it.

PR #417, opened 2018-10-30
Add link to start and finish launch responses

ihar-kahadouski:


src/main/java/com/epam/ta/reportportal/ws/controller/LaunchController.java
```diff
@@ -310,4 +316,9 @@ public OperationCompletionRS importLaunch(@PathVariable String projectName, @Req
 		return importLaunchHandler.importLaunch(extractProjectDetails(user, normalizeId(projectName)), user, "XUNIT", file);
 	}
 
+	private String generateLaunchLink(String host, String projectName, String id) {
+		StringBuilder sb = new StringBuilder("http://");
```
pbortnik: what if url has "https" protocol?

PR #434, opened 2018-11-14
Launch import performance improvement by transactions refactoring

Yumfriez:


src/main/java/com/epam/ta/reportportal/core/imprt/impl/junit/XunitParseJob.java
```diff
@@ -40,6 +42,7 @@
 
 	private InputStream xmlInputStream;
 
+	@Transactional(isolation = Isolation.READ_UNCOMMITTED)
```
pbortnik: check if we really need read uncommited?

PR #441, opened 2018-11-21
Retries api impl

Yumfriez:


src/main/java/com/epam/ta/reportportal/core/item/impl/StartTestItemHandlerImpl.java
```diff
@@ -115,6 +115,9 @@ public ItemCreatedRS startChildItem(ReportPortalUser user, ReportPortalUser.Proj
 		if (null == item.getUniqueId()) {
 			item.setUniqueId(identifierGenerator.generate(item, launch));
 		}
+		if(rq.isRetry() != null && rq.isRetry()) {
```
pbortnik: Boolean utils is true

src/main/java/com/epam/ta/reportportal/core/launch/impl/FinishLaunchHandler.java
```diff
@@ -110,6 +110,8 @@ public Launch finishLaunch(Long launchId, FinishExecutionRQ finishLaunchRQ, Repo
 		}
 		launch.setStatus(statusEnum.orElse(fromStatisticsStatus));
 
+		testItemRepository.handleRetriesStatistics(launch.getId());
```
pbortnik: let's move this logic into "after launch finished handler"

PR #442, opened 2018-11-21
Delete launches detailed response

ihar-kahadouski:


src/main/java/com/epam/ta/reportportal/core/launch/impl/DeleteLaunchHandler.java
```diff
@@ -86,18 +90,41 @@ public OperationCompletionRS deleteLaunch(Long launchId, ReportPortalUser.Projec
...
-		expect(launches.size(), not(equalTo(0))).verify(BAD_REQUEST_ERROR, "ids=" + Arrays.toString(ids));
-		launches.forEach(l -> validate(l, user, projectDetails));
+	//TODO Analyzer
+	public DeleteLaunchesRS deleteLaunches(Long[] ids, ReportPortalUser.ProjectDetails projectDetails, ReportPortalUser user) {
+		List<Long> notFound = Lists.newArrayList();
+		List<ReportPortalException> exceptions = Lists.newArrayList();
+		List<Launch> toDelete = Lists.newArrayList();
+
+		stream(ids).forEach(id -> {
+			Optional<Launch> optionalLaunch = launchRepository.findById(id);
```
pbortnik: orElse(it -> notFound.add(id))
so if-else could be removed

ihar-kahadouski: public T orElse(T other)
so orElse doesn't work that way

PR #454, opened 2018-11-28
 [EPMRPP-34008] Changing test item status 

ihar-kahadouski:


src/main/java/com/epam/ta/reportportal/core/item/impl/FinishTestItemHandlerImpl.java
```diff
@@ -139,13 +139,15 @@ private TestItemResults processItemResults(Long projectId, TestItem testItem, Fi
...
+					issueEntity.setIssueId(testItem.getItemId());
+					testItemResults.setIssue(issueEntity);
 				}
 			} else {
 				IssueType toInvestigate = issueTypeHandler.defineIssueType(testItem.getItemId(), projectId, TO_INVESTIGATE.getLocator());
 				issueEntity.setIssueType(toInvestigate);
+				issueEntity.setIssueId(testItem.getItemId());
+				testItemResults.setIssue(issueEntity);
 			}
-			issueEntity.setIssueId(testItem.getItemId());
```
pbortnik: why did you moved it into if and else structure?
it should always be called

src/main/java/com/epam/ta/reportportal/core/item/impl/UpdateTestItemHandlerImpl.java
```diff
@@ -173,8 +173,7 @@ public OperationCompletionRS updateTestItem(ReportPortalUser.ProjectDetails proj
 		Optional<StatusEnum> statusEnum = StatusEnum.fromValue(rq.getStatus());
 
 		if (statusEnum.isPresent()) {
-			testItem.getItemResults().setIssue(null);
-			testItem.getItemResults().setStatus(statusEnum.get());
+			changeStatus(testItem, statusEnum.get(), user, projectDetails);
```
pbortnik: let's move the logic of status changing into different handler
for example ItemStatusChangeHandler

src/main/java/com/epam/ta/reportportal/core/item/impl/UpdateTestItemHandlerImpl.java
```diff
@@ -353,4 +352,179 @@ private void verifyTestItem(TestItem item, Long id) throws BusinessRuleViolation
...
+		messageBus.publishActivity(new TestItemStatusChangedEvent(before, TO_ACTIVITY_RESOURCE.apply(testItem), userId));
+
+		if (PASSED.equals(providedStatus)) {
+			changeStatusRecursively(testItem, userId);
+		}
+	}
+
+	private void changeStatusFromPassed(TestItem testItem, StatusEnum providedStatus, Long userId, Long projectId) {
+		expect(providedStatus, statusIn(SKIPPED, FAILED)).verify(INCORRECT_REQUEST,
+				"actual status: " + testItem.getItemResults().getStatus() + " can not be changed to: " + providedStatus
```
pbortnik: Let's say in what way status can be changed.
'Status 'passed' can be switched only to skipped or failed'
and in other places also in such way

src/main/java/com/epam/ta/reportportal/core/item/impl/UpdateTestItemHandlerImpl.java
```diff
@@ -353,4 +352,179 @@ private void verifyTestItem(TestItem item, Long id) throws BusinessRuleViolation
...
+		Optional<ItemAttribute> skippedIssueAttribute = itemAttributeRepository.findSystemAttributeByLaunchIdAndValue(testItem.getLaunch()
+				.getId(), SKIPPED_ISSUE_KEY);
+
+		testItem.getItemResults().setStatus(providedStatus);
+		if (FAILED.equals(providedStatus) || (SKIPPED.equals(providedStatus) && skippedIssueAttribute.isPresent()
+				&& skippedIssueAttribute.get().getValue().equals("true"))) {
+			setToInvestigateIssue(testItem, projectId);
+		}
+
+		testItemRepository.save(testItem);
```
pbortnik: we don't need to say this. as it is in transaction it will be saved after transaction commit

src/main/java/com/epam/ta/reportportal/core/item/impl/UpdateTestItemHandlerImpl.java
```diff
@@ -353,4 +352,179 @@ private void verifyTestItem(TestItem item, Long id) throws BusinessRuleViolation
...
+		}
+	}
+
+	private void changeParentsStatusesToFailed(TestItem testItem, StatusEnum oldParentStatus, Long userId) {
+		if (!oldParentStatus.equals(StatusEnum.FAILED)) {
+			TestItem parent = testItem.getParent();
+			TestItemActivityResource before = TO_ACTIVITY_RESOURCE.apply(parent);
+			while (parent != null) {
+				parent.getItemResults().setStatus(StatusEnum.FAILED);
+				testItemRepository.save(parent);
```
pbortnik: it is in transaction

src/main/java/com/epam/ta/reportportal/core/item/impl/FinishTestItemHandlerImpl.java
ihar-kahadouski: if we finish item with status skipped -> to item results sets empty issue only with id -> hibernate trying to save it and throws constraint violation exception

PR #459, opened 2018-12-03
[EPMRPP-37281] Swagger sorting field description fix

ihar-kahadouski:


src/main/java/com/epam/ta/reportportal/core/configs/Swagger2Configuration.java
```diff
@@ -183,7 +183,7 @@ public void apply(OperationContext context) {
 							.name("page.sort")
 							.modelRef(stringModel)
 							.allowMultiple(true)
-							.description("Sorting criteria in the format: property(,asc|desc). " + "Default sort order is ascending. "
+							.description("Sorting criteria in the format: property, (asc|desc). " + "Default sort order is ascending. "
```
pbortnik: it is not a mistake. it is a format how user should send sort criteria

PR #485, opened 2018-12-11
[EPMRPP-35695]  Add validation if user filter belongs project 

ihar-kahadouski:


src/main/java/com/epam/ta/reportportal/core/filter/impl/DeleteUserFilterHandlerImpl.java
```diff
@@ -20,37 +20,46 @@
...
+			ReportPortalAclHandler aclHandler) {
 		this.userFilterRepository = userFilterRepository;
 		this.messageBus = messageBus;
-		this.getFilterHandler = getFilterHandler;
+		this.aclHandler = aclHandler;
 	}
 
 	@Override
 	public OperationCompletionRS deleteFilter(Long id, ReportPortalUser.ProjectDetails projectDetails, ReportPortalUser user) {
-		UserFilter userFilter = getFilterHandler.getFilter(id, projectDetails, user);
```
pbortnik: you can't replace this with just find by id case only user with ADMINISTRATION acl permission could delete filters

PR #494, opened 2018-12-14
[EPMRPP-37405] Fix updating item attributes

ihar-kahadouski:


src/main/java/com/epam/ta/reportportal/ws/converter/builders/LaunchBuilder.java
```diff
@@ -127,4 +128,13 @@ public LaunchBuilder addEndTime(Date date) {
 	public Launch get() {
 		return launch;
 	}
+
+	private ItemAttribute fromResource(ItemAttributeResource val) {
```
pbortnik: make it as a function in converters

PR #528, opened 2019-01-04
[EPMRPP-37599]  Do not update attributes, if they are null in request 

ihar-kahadouski:


src/main/java/com/epam/ta/reportportal/ws/converter/builders/LaunchBuilder.java
```diff
@@ -101,16 +101,18 @@ public LaunchBuilder addAttributes(Set<ItemAttributeResource> attributes) {
...
-				.filter(ItemAttribute::isSystem)
-				.collect(Collectors.toSet()));
-		ofNullable(attributes).ifPresent(it -> it.stream().map(val -> {
-			ItemAttribute itemAttribute = FROM_RESOURCE.apply(val);
-			itemAttribute.setLaunch(launch);
-			return itemAttribute;
-		}).forEach(overwrittenAttributes::add));
-		launch.setAttributes(overwrittenAttributes);
+		if (attributes != null) {
+			final HashSet<ItemAttribute> overwrittenAttributes = Sets.newHashSet(launch.getAttributes()
```
pbortnik: do we need to directly say that it is HashSet? can we use set that is created after stream collect and not to create a new one?

src/main/java/com/epam/ta/reportportal/ws/converter/builders/TestItemBuilder.java
```diff
@@ -113,16 +113,18 @@ public TestItemBuilder addAttributes(Set<ItemAttributeResource> attributes) {
...
-				.filter(ItemAttribute::isSystem)
-				.collect(Collectors.toSet()));
-		ofNullable(attributes).ifPresent(it -> it.stream().map(val -> {
-			ItemAttribute itemAttribute = FROM_RESOURCE.apply(val);
-			itemAttribute.setTestItem(testItem);
-			return itemAttribute;
-		}).forEach(overwrittenAttributes::add));
-		testItem.setAttributes(overwrittenAttributes);
+		if (attributes != null) {
+			final HashSet<ItemAttribute> overwrittenAttributes = Sets.newHashSet(testItem.getAttributes()
```
pbortnik: same as above

PR #545, opened 2019-01-17
EPMRPP-37614 || Cumulative widget impl with widget options refactoring

Yumfriez:


src/main/java/com/epam/ta/reportportal/core/widget/util/WidgetOptionUtil.java
```diff
@@ -37,29 +39,31 @@ private WidgetOptionUtil() {
...
+		Optional<Object> value = ofNullable(widgetOptions).map(wo -> ofNullable(wo.getOptions()).map(options -> options.get(key))
+				.orElse(null));
+
+		value.ifPresent(v -> expect(v, String.class::isInstance).verify(
 				ErrorType.OBJECT_RETRIEVAL_ERROR,
 				Suppliers.formattedSupplier("Wrong widget option value type for key = '{}'. String expected.", key)
 		));
 
-		return (String) value;
+		return (String) value.orElse(null);
```
pbortnik: let's it better be value.get() cause you have validated it

Yumfriez: decided to leave *.orElse(null) statement

PR #551, opened 2019-01-22
[EPMRPP-37765] Set default mode to launch if not present in request

ihar-kahadouski:


src/main/java/com/epam/ta/reportportal/ws/converter/builders/LaunchBuilder.java
```diff
@@ -62,7 +62,8 @@ public LaunchBuilder addStartRQ(StartLaunchRQ request) {
 		launch.setStatus(StatusEnum.IN_PROGRESS);
 		launch.setUuid(Optional.ofNullable(request.getUuid()).orElse(UUID.randomUUID().toString()));
 		addDescription(request.getDescription());
-		ofNullable(request.getMode()).ifPresent(it -> launch.setMode(LaunchModeEnum.valueOf(request.getMode().name())));
+		launch.setMode(LaunchModeEnum.findByName(ofNullable(request.getMode()).map(Enum::name).orElse(LaunchModeEnum.DEFAULT.name()))
```
pbortnik: result of inner ofnullable will alwayse exists
let's better make it using ifpresent to avoid dublicating orElse statement

PR #589, opened 2019-02-11
Default project fix

Yumfriez:


src/main/java/com/epam/ta/reportportal/core/user/impl/CreateUserHandlerImpl.java
```diff
@@ -262,6 +265,7 @@ public CreateUserRS createUser(CreateUserRQConfirm request, String uuid) {
 
 		try {
 			userRepository.save(newUser);
+			newUser.setDefaultProject(defaultProject);
```
pbortnik: default project is a project that user is assigned while creating bid
personal project - is a default project for user

why this logic has been changed ?
probably we should remove a default project from user and leave it only in the list of projects

pbortnik: as we discussed - we are removing default project from database at all

PR #600, opened 2019-02-18
Jira and Rally integrations creation logic update

Yumfriez:


src/main/java/com/epam/ta/reportportal/core/integration/util/AbstractBtsIntegrationService.java
```diff
@@ -0,0 +1,139 @@
...
+
+import static com.epam.ta.reportportal.commons.Predicates.isPresent;
+import static com.epam.ta.reportportal.commons.Predicates.not;
+import static com.epam.ta.reportportal.commons.validation.BusinessRule.expect;
+
+/**
+ * @author <a href="mailto:ivan_budayeu@epam.com">Ivan Budayeu</a>
+ */
+@Service
+public abstract class AbstractBtsIntegrationService implements IntegrationService {
```
pbortnik: what do you think about covering new services with tests?

PR #601, opened 2019-02-18
EPMRPP-37698 || Notifications refactoring api

Yumfriez:


src/main/java/com/epam/ta/reportportal/core/project/impl/UpdateProjectHandlerImpl.java
```diff
@@ -104,6 +106,9 @@
 
 	private final ProjectConverter projectConverter;
 
+	@Autowired
+	private ObjectMapper objectMapper;
```
pbortnik: can't find the usage of objectMapper here. why do we need it?

Yumfriez: removed

PR #602, opened 2019-02-19
EPMRPP-38048 || Unique filter validation

Yumfriez:


src/main/java/com/epam/ta/reportportal/core/filter/impl/UpdateUserFilterHandlerImpl.java
```diff
@@ -103,6 +103,16 @@ public OperationCompletionRS updateUserFilter(Long userFilterId, UpdateUserFilte
 				projectDetails.getProjectId(),
 				user.getUserId()
 		);
+
+		if (!userFilter.getName().equals(updateRQ.getName())) {
```
pbortnik: test for this logic

Yumfriez: done

PR #605, opened 2019-02-20
Copy scripts from GitHub

ihar-kahadouski:


build.gradle
```diff
@@ -243,6 +243,27 @@ bootJar {
     archiveName = "${baseName}.${extension}"
 }
 
+private static void copyFromUrl(GString localScriptPath, String scriptUrl) {
```
pbortnik: if i run it locally, should i use gradle to get scripts? probably we can try to make it in java code?

build.gradle
```diff
@@ -243,6 +243,27 @@ bootJar {
...
+    def file = new File(localScriptPath)
+    if (file.exists()) {
+        file.delete()
+    }
+    file.createNewFile()
+    file << new URL(scriptUrl).openStream()
+}
+
+task copyScripts {
```
pbortnik: copyTestDatabaseScripts

PR #609, opened 2019-02-21
Jobs unit testing

ihar-kahadouski:


src/main/java/com/epam/ta/reportportal/job/LogCleanerServiceImpl.java
```diff
@@ -85,19 +85,21 @@ public void removeOutdatedLogs(Project project, Duration period, AtomicLong remo
 		try (Stream<Long> launchIds = launchRepository.streamIdsModifiedBefore(project.getId(), TO_LOCAL_DATE_TIME.apply(endDate))) {
 			launchIds.forEach(id -> {
 				try (Stream<Long> ids = testItemRepository.streamTestItemIdsByLaunchId(id)) {
-					ids.forEach(itemId -> {
+					List<Long> logIds = ids.peek(itemId -> {
```
pbortnik: it's incorrect. you apply peek to list of item ids and then collect it assuming that you will get the list of log ids but you will not. now you just collected stream of item ids into list of item ids.
you probably need to use map with returning log values if you need them
or if you need test items - just rename.

src/main/java/com/epam/ta/reportportal/job/LogCleanerServiceImpl.java
```diff
@@ -85,19 +85,21 @@ public void removeOutdatedLogs(Project project, Duration period, AtomicLong remo
...
 						removeAttachmentsOfLogs(logs, attachmentsCount, thumbnailsCount);
-					});
-					long count = logRepository.deleteByPeriodAndTestItemIds(period, ids.collect(Collectors.toList()));
+					}).collect(Collectors.toList());
+					long count = logRepository.deleteByPeriodAndTestItemIds(period, logIds);
 					removedLogsCount.addAndGet(count);
 					removedLogsInThreadCount.addAndGet(count);
 				} catch (Exception e) {
 					//do nothing
+					LOGGER.error("Error during cleaning outdated logs {}", e);
```
pbortnik: if you have added something -> remove do nothing comment

PR #624, opened 2019-03-04
[EPMRPP-38158] Widget content loader tests

ihar-kahadouski:


Jenkinsfile
```diff
@@ -1,3 +1,19 @@
+/*
```
pbortnik: Copyright in jenkinsfile breaks build in jenkins.
Remove it. Probably find a way how to update copyright at only modified files.

PR #633, opened 2019-03-11
EPMRPP-38220 || Admin role self-update fix

Yumfriez:


src/main/java/com/epam/ta/reportportal/core/user/EditUserHandler.java
```diff
@@ -36,11 +35,11 @@
 	 *
 	 * @param username
 	 * @param editUserRQ
-	 * @param role
+	 * @param updater
 	 * @return
 	 * @throws ReportPortalException
 	 */
-	OperationCompletionRS editUser(String username, EditUserRQ editUserRQ, UserRole role);
+	OperationCompletionRS editUser(String username, EditUserRQ editUserRQ, ReportPortalUser updater);
```
pbortnik: maybe 'editor' is a better word to wire method name and parameter?

Yumfriez: renamed

PR #657, opened 2019-03-20
Search project names endpoint impl

Yumfriez:


src/main/java/com/epam/ta/reportportal/ws/controller/ProjectController.java
```diff
@@ -337,6 +337,16 @@ public ProjectInfoResource getProjectInfo(@PathVariable String projectName,
 		return getProjectHandler.getAllProjectNames();
 	}
 
+	@Transactional(readOnly = true)
+	@PreAuthorize(ADMIN_ONLY)
+	@RequestMapping(value = "/names/search", method = RequestMethod.GET)
```
pbortnik: @GetMapping

src/main/java/com/epam/ta/reportportal/ws/controller/ProjectController.java
```diff
@@ -337,6 +337,16 @@ public ProjectInfoResource getProjectInfo(@PathVariable String projectName,
...
+	@ResponseBody
```
pbortnik: it is @RestController so we don't need this annotation

src/main/java/com/epam/ta/reportportal/ws/controller/ProjectController.java
Yumfriez: fixed

src/main/java/com/epam/ta/reportportal/ws/controller/ProjectController.java
Yumfriez: fixed

PR #665, opened 2019-04-01
Not-null validation for description has been removed

Yumfriez:


src/test/java/com/epam/ta/reportportal/core/project/settings/impl/CreateProjectSettingsHandlerImplTest.java
```diff
@@ -111,7 +111,7 @@ void maxSubtypesCount() {
 
 	private Set<ProjectIssueType> getSubTypes() {
 		HashSet<ProjectIssueType> subTypes = new HashSet<>();
-		for (int i = 1; i < 12; i++) {
+		for (int i = 1; i < 16; i++) {
```
pbortnik: Are you a programmer ?? Do you count from 1 and not from 0????

PR #669, opened 2019-04-02
Data source init properties fix

Yumfriez:


src/main/resources/application.yaml
```diff
@@ -70,8 +66,16 @@ rp:
...
+    type: com.zaxxer.hikari.HikariDataSource
+    driverClassName: org.postgresql.Driver
+    jdbcUrl: ${rp.db.url}
+    username: ${rp.db.user}
+    password: ${rp.db.pass}
+    maximumPoolSize: 27
+
   db:
-    url: jdbc:postgresql://${rp.db.host}:${rp.db.port}/${rp.db.name}?user=${rp.db.user}&password=${rp.db.pass}
+    url: jdbc:postgresql://${rp.db.host}:${rp.db.port}/${rp.db.name}
```
pbortnik: why did you remove user and password from here?

Yumfriez: This statement was split into 3 key-value statements:
url: url
username: username
password: password

PR #673, opened 2019-04-02
EPMRPP-35500 || Statistics update api

Yumfriez:


src/main/java/com/epam/ta/reportportal/core/item/impl/FinishTestItemHandlerImpl.java
```diff
@@ -120,7 +121,7 @@ private TestItemResults processItemResults(Long projectId, TestItem testItem, Fi
 		}
 
 		if (Preconditions.statusIn(FAILED, SKIPPED).test(testItemResults.getStatus()) && !hasChildren
-				&& !ofNullable(testItem.getRetryOf()).isPresent()) {
+				&& !ofNullable(testItem.getRetryOf()).isPresent() && testItem.getType() != TestItemTypeEnum.SUITE) {
```
pbortnik: It has become pretty complicated, so we can to create a separate Precondition as a static fields for example which would have these conditions.
CAN_HAVE_ISSUE = ...

Yumfriez: moved to private method

PR #678, opened 2019-04-05
EPMRPP-38208 || Binary data saving fix

Yumfriez:


src/main/java/com/epam/ta/reportportal/core/log/impl/AsyncCreateLogHandler.java
```diff
@@ -60,27 +59,24 @@
...
 
 		Log log = new LogBuilder().addSaveLogRq(createLogRQ).addTestItem(testItem).get();
-		try {
-			logRepository.save(log);
-		} catch (Exception exc) {
-			throw new ReportPortalException("Error while Log instance creating.", exc);
-		}
+		logRepository.save(log);
+
 		if (null != file) {
```
ihar-kahadouski: What if file not null, but content is empty?

Yumfriez: If file exists we should save it, cause it can be empty just to 'mark' any log

PR #680, opened 2019-04-08
Saucelabs integration

pbortnik:


src/main/java/com/epam/ta/reportportal/core/integration/GetIntegrationHandler.java
```diff
@@ -42,4 +43,12 @@
...
 
 	Integration getEnabledBtsIntegration(ReportPortalUser.ProjectDetails projectDetails, Long integrationId);
+
+	List<IntegrationResource> getGlobalIntegrations();
+
+	List<IntegrationResource> getGlobalIntegrations(String pluginName);
+
+	List<IntegrationResource> getProjectIntegrations(ReportPortalUser.ProjectDetails projectDetails);
+
+	List<IntegrationResource> getProjectIntegrations(String pluginName, ReportPortalUser.ProjectDetails projectDetails);
```
Yumfriez: java docs

src/main/java/com/epam/ta/reportportal/core/integration/plugin/impl/CreatePluginHandlerImpl.java
```diff
@@ -166,38 +168,13 @@ private void createTempPluginsFolderIfNotExists(Path path) {
...
-			);
-		}
-
-		return reportPortalIntegration.get();
-	}
-
-	private IntegrationType retrieveIntegrationType(PluginInfo pluginInfo, ReportPortalIntegrationEnum reportPortalIntegration) {
+	private IntegrationType retrieveIntegrationType(PluginInfo pluginInfo) {
 
 		IntegrationType integrationType = integrationTypeRepository.findByName(pluginInfo.getId()).map(it -> {
```
Yumfriez: Here should be a validation that will prevent 'email' and 'ldap' retrieving

src/main/java/com/epam/ta/reportportal/core/integration/plugin/impl/CreatePluginHandlerImpl.java
```diff
@@ -253,6 +230,15 @@ private IntegrationType startUpPlugin(String newPluginId, Optional<PluginWrapper
 
 			integrationType.setName(newLoadedPluginId.get());
 
+			Optional<ReportPortalExtensionPoint> instance = pluginBox.getInstance(newLoadedPluginId.get(),
+					ReportPortalExtensionPoint.class
+			);
+
+			if (instance.isPresent()) {
+				IntegrationDetailsProperties.COMMANDS.setValue(integrationType.getDetails(), instance.get().getCommandNames());
+				integrationType.setIntegrationGroup(IntegrationGroupEnum.valueOf(instance.get().getIntegrationGroup().name()));
```
Yumfriez: is it really necessary to get Integration group name and then create a enum from it?

src/main/java/com/epam/ta/reportportal/core/integration/util/IntegrationService.java
```diff
@@ -26,42 +26,7 @@
...
 
-	/**
-	 * Create {@link Integration} with {@link Integration#project == NULL}
-	 *
-	 * @param integrationTypeName {@link com.epam.ta.reportportal.entity.integration.IntegrationType#name}
-	 * @param integrationParams   {@link com.epam.ta.reportportal.entity.integration.IntegrationParams#params}
-	 * @return new {@link Integration}
-	 */
-	Integration createGlobalIntegration(String integrationTypeName, Map<String, Object> integrationParams);
+	Map<String, Object> retrieveIntegrationParams(Map<String, Object> integrationParams);
```
Yumfriez: java docs

src/main/java/com/epam/ta/reportportal/core/integration/util/IntegrationService.java
```diff
@@ -26,42 +26,7 @@
...
-	/**
-	 * Updated {@link Integration} for {@link com.epam.ta.reportportal.entity.project.Project} with provided ID
-	 *
-	 * @param id                {@link Integration#id}
-	 * @param projectDetails    {@link com.epam.ta.reportportal.commons.ReportPortalUser.ProjectDetails}
-	 * @param integrationParams {@link com.epam.ta.reportportal.entity.integration.IntegrationParams#params}
-	 * @return updated {@link Integration}
-	 */
-	Integration updateProjectIntegration(Long id, ReportPortalUser.ProjectDetails projectDetails, Map<String, Object> integrationParams);
+	boolean validateIntegration(Integration integration, ReportPortalUser.ProjectDetails projectDetails);
```
Yumfriez: Maybe this method can be split into two:
1) for global integration
2) for project integration
If we have project-/global-scoped integrations, we can split validation logic according to this difference, so if we want to validate global integration, we can just call appropriate method without providing 'NULL' for ProjectDetails argument

src/main/java/com/epam/ta/reportportal/core/integration/util/property/IntegrationDetailsProperties.java
```diff
@@ -45,7 +46,7 @@
 		return ofNullable(details.get(this.attribute)).map(String::valueOf);
```
Yumfriez: If here can be types other than 'String', it would be more correct to return 'Object' instead of String representation.
Or 'objectMapper' can be added to the 'getValue' args and any 'Object' will be serialized to the 'String' and then will be deserialized.

src/main/java/com/epam/ta/reportportal/core/integration/util/property/ReportPortalIntegrationEnum.java
```diff
@@ -1,54 +1,54 @@
...
+//
+//import com.epam.ta.reportportal.entity.enums.IntegrationGroupEnum;
+//
+//import java.util.Arrays;
+//import java.util.Optional;
+//
+///**
+// * @author <a href="mailto:ivan_budayeu@epam.com">Ivan Budayeu</a>
+// */
+//public enum ReportPortalIntegrationEnum {
```
Yumfriez: If we don't use it, it should be removed

src/main/java/com/epam/ta/reportportal/ws/controller/FileStorageController.java
```diff
@@ -101,7 +101,7 @@ public OperationCompletionRS deletePhoto(@AuthenticationPrincipal ReportPortalUs
 	 * @param response    Response
 	 * @param inputStream Stored data
 	 */
-	private void toResponse(HttpServletResponse response, InputStream inputStream) {
+	public static void toResponse(HttpServletResponse response, InputStream inputStream) {
```
Yumfriez: It can be moved to the separate util class

src/main/java/com/epam/ta/reportportal/ws/converter/converters/TicetConverter.java
```diff
@@ -25,9 +25,9 @@
 /**
  * @author Pavel Bortnik
  */
-public final class IntegrationIssueConverter {
+public final class TicetConverter {
```
Yumfriez: Ticket*

src/main/resources/application.yaml
```diff
@@ -88,7 +88,7 @@ rp:
   analytics:
     enableByDefault: true
   plugins:
-    path: /plugins
+    path: /Users/pavel_bortnik/IdeaProjects/service-api/plugins
```
Yumfriez: =)

src/main/resources/datastore.properties
```diff
@@ -13,7 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-datastore.default.path=${rp.binarystore.path:/data/storage}
+datastore.default.path=${rp.binarystore.path:/Users/pavel_bortnik/IdeaProjects/service-api/data/storage}
```
Yumfriez: change it back

src/main/java/com/epam/ta/reportportal/core/integration/plugin/impl/CreatePluginHandlerImpl.java
pbortnik: different enums. on is in dao, another is in plugin api

PR #729, opened 2019-05-04
[EPMRPP-38819] Update reporting

ihar-kahadouski:


src/main/java/com/epam/ta/reportportal/core/launch/TerminateLaunchHandler.java
```diff
@@ -0,0 +1,104 @@
...
+import static com.epam.ta.reportportal.commons.validation.BusinessRule.expect;
+import static com.epam.ta.reportportal.commons.validation.Suppliers.formattedSupplier;
+import static com.epam.ta.reportportal.entity.enums.StatusEnum.*;
+import static com.epam.ta.reportportal.entity.project.ProjectRole.PROJECT_MANAGER;
+import static com.epam.ta.reportportal.ws.model.ErrorType.*;
+
+/**
+ * @author <a href="mailto:ihar_kahadouski@epam.com">Ihar Kahadouski</a>
+ */
+public interface TerminateLaunchHandler {
```
pbortnik: it contains only validation methods, so it is probably better to rename it

ihar-kahadouski: Provide please your alternative

pbortnik: LaunchValidator
and it is probably better to make it as an util class ? 
it's not really interface as it is

ihar-kahadouski: fixed

PR #741, opened 2019-05-15
EPMRPP-38865 || Pattern analysis activities api

Yumfriez:


src/main/resources/application.yaml
```diff
@@ -72,7 +72,7 @@ rp:
   db:
     url: jdbc:postgresql://${rp.db.host}:${rp.db.port}/${rp.db.name}
     name: reportportal
-    host: postgres
+    host: localhost
```
pbortnik: change back

src/main/resources/datastore.properties
```diff
@@ -13,7 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-datastore.default.path=${rp.binarystore.path:/data/storage}
+datastore.default.path=${rp.binarystore.path:/Users/yumfriez/data/storage}
```
pbortnik: change back

Yumfriez: fixed

src/main/resources/application.yaml
Yumfriez: fixed

PR #779, opened 2019-06-06
Pattern matched activities for item details page impl

Yumfriez:


src/main/java/com/epam/ta/reportportal/core/activity/impl/ActivityHandlerImpl.java
```diff
@@ -106,10 +107,19 @@ public ActivityResource getActivity(ReportPortalUser.ProjectDetails projectDetai
 		);
 
 		Sort sortByCreationDateDesc = Sort.by(Sort.Direction.DESC, CRITERIA_CREATION_DATE);
+
+		Filter patternActivityFilter = Filter.builder()
```
pbortnik: move it to a different method, 'buildPatternActivityFilter'
and give it documentation

Yumfriez: updated

PR #802, opened 2019-06-20
Creation date ordering for plugins impl

Yumfriez:


src/main/java/com/epam/ta/reportportal/core/integration/plugin/impl/GetPluginHandlerImpl.java
```diff
@@ -41,6 +42,10 @@ public GetPluginHandlerImpl(IntegrationTypeRepository integrationTypeRepository)
 
 	@Override
 	public List<IntegrationTypeResource> getPlugins() {
-		return integrationTypeRepository.findAll().stream().map(IntegrationTypeConverter.TO_RESOURCE).collect(Collectors.toList());
+		return integrationTypeRepository.findAll()
```
pbortnik: maybe it's better to do this work at database level?

PR #805, opened 2019-06-24
[EPMRPP-39503] Send notification with updated data in case auto analysis enabled

ihar-kahadouski:


src/main/java/com/epam/ta/reportportal/core/events/handler/LaunchFinishedEventHandler.java
```diff
@@ -133,8 +133,8 @@ public void onApplicationEvent(LaunchFinishedEvent event) throws ExecutionExcept
 		if (BooleanUtils.isTrue(analyzerConfig.getIsAutoAnalyzerEnabled()) && analyzerServiceAsync.hasAnalyzers()) {
 			List<Long> itemIds = analyzeCollectorFactory.getCollector(AnalyzeItemsMode.TO_INVESTIGATE)
 					.collectItems(project.getId(), launch.getId(), null);
-			analyzerServiceAsync.analyze(launch, itemIds, analyzerConfig)
-					.thenApply(it -> logIndexer.indexItemsLogs(project.getId(), launch.getId(), itemIds, analyzerConfig));
+			analyzerServiceAsync.analyze(launch, itemIds, analyzerConfig).get();
```
pbortnik: Check why it is still working in one thread with main finish launch logic

ihar-kahadouski: because of transactions

PR #813, opened 2019-06-27
move rerun handling into service

ihar-kahadouski:


src/main/java/com/epam/ta/reportportal/core/item/impl/StartTestItemHandlerImpl.java
```diff
@@ -82,19 +84,15 @@ public ItemCreatedRS startRootItem(ReportPortalUser user, ReportPortalUser.Proje
...
-
-		//pzdc. need refactoring
-		if (launch.isRerun() && (itemOptional = testItemRepository.findByNameAndLaunchWithoutParents(rq.getName(),
-				launch.getId()
-		)).isPresent()) {
-			item = handleRerun(rq, launch, itemOptional.get(), null);
-		} else {
-			testItemRepository.save(item);
-			generateUniqueId(launch, item, String.valueOf(item.getItemId()));
+		Optional<ItemCreatedRS> rerun = rerunHandler.handleTestItem(rq, launch, null);
```
pbortnik: it is better to give it a better name
rerunCreatedRs or something

pbortnik: Also, maybe it is a good idea to call it only if rq has "rerun flag"

pbortnik: Also, maybe it is good to create separate methods for handleRootItem and children
cause from the context it is hard to recognize what does this 'null' mean here

ihar-kahadouski: fixed

PR #816, opened 2019-07-02
Fix analyzer on update item

ihar-kahadouski:


src/main/java/com/epam/ta/reportportal/core/launch/impl/UpdateLaunchHandlerImpl.java
```diff
@@ -162,7 +163,11 @@ public OperationCompletionRS startLaunchAnalyzer(ReportPortalUser.ProjectDetails
 		List<Long> itemIds = collectItemsByModes(project, user.getUsername(), launch.getId(), analyzeRQ.getAnalyzeItemsMode());
 
 		analyzerServiceAsync.analyze(launch, itemIds, analyzerConfig)
-				.thenApply(it -> logIndexer.indexItemsLogs(project.getId(), launch.getId(), itemIds, analyzerConfig));
+				.thenApply(it -> CompletableFuture.supplyAsync(() -> logIndexer.indexItemsLogs(project.getId(),
```
pbortnik: i think we don't need indexing to be async here, cause it is already async for analyze

src/main/java/com/epam/ta/reportportal/core/events/handler/TestItemFinishedEventHandler.java
```diff
@@ -53,11 +55,11 @@ public void onApplicationEvent(ItemFinishedEvent itemFinishedEvent) {
 
 		AnalyzerConfig analyzerConfig = AnalyzerUtils.getAnalyzerConfig(project);
 
-		logIndexer.indexItemsLogs(
+		CompletableFuture.supplyAsync(() -> logIndexer.indexItemsLogs(
```
pbortnik: maybe we need the whole event async here that is executed after main transaction is commited

src/main/java/com/epam/ta/reportportal/core/events/handler/LaunchFinishedEventHandler.java
```diff
@@ -127,16 +127,18 @@ public void onApplicationEvent(LaunchFinishedEvent event) throws ExecutionExcept
...
 		if (BooleanUtils.isTrue(analyzerConfig.getIsAutoAnalyzerEnabled()) && analyzerServiceAsync.hasAnalyzers()) {
 			List<Long> itemIds = analyzeCollectorFactory.getCollector(AnalyzeItemsMode.TO_INVESTIGATE)
 					.collectItems(project.getId(), launch.getId(), null);
-			analyzerServiceAsync.analyze(launch, itemIds, analyzerConfig)
-					.thenApply(it -> logIndexer.indexItemsLogs(project.getId(), launch.getId(), itemIds, analyzerConfig));
+			logIndexer.indexLaunchLogs(project.getId(), launch.getId(), analyzerConfig).join();
+			analyzerServiceAsync.analyze(launch, itemIds, analyzerConfig).join();
+			CompletableFuture.supplyAsync(() -> logIndexer.indexItemsLogs(project.getId(), launch.getId(), itemIds, analyzerConfig));
+		} else {
+			logIndexer.indexLaunchLogs(project.getId(), launch.getId(), analyzerConfig);
```
pbortnik: check the if condition
you will start indexing even you don't have any analyzers

ihar-kahadouski: logs should be indexed in any case. It doesn't depend on auto analysis

src/main/java/com/epam/ta/reportportal/core/events/handler/TestItemFinishedEventHandler.java
ihar-kahadouski: fixed

src/main/java/com/epam/ta/reportportal/core/launch/impl/UpdateLaunchHandlerImpl.java
ihar-kahadouski: fixed

PR #818, opened 2019-07-04
Logs exclusion condition fix

Yumfriez:


src/main/java/com/epam/ta/reportportal/core/log/impl/GetLogHandlerImpl.java
```diff
@@ -143,4 +149,19 @@ private Log findAndValidate(Long logId, ReportPortalUser.ProjectDetails projectD
 
 		return log;
 	}
+
+	private boolean isLogsExclusionRequired(Long parentId, Map<String, String> params) {
```
pbortnik: give it a clear documentation

Yumfriez: added

PR #833, opened 2019-07-11
[Fix] Launch link refactoring

ihar-kahadouski:


src/main/java/com/epam/ta/reportportal/ws/controller/LaunchAsyncController.java
```diff
@@ -77,11 +77,10 @@ public StartLaunchRS startLaunch(@PathVariable String projectName,
 	public FinishLaunchRS finishLaunch(@PathVariable String projectName, @PathVariable String launchId,
 			@RequestBody @Validated FinishExecutionRQ finishLaunchRQ, @AuthenticationPrincipal ReportPortalUser user,
 			HttpServletRequest request) {
+		finishLaunchRQ.setBaseURL(composeBaseUrl(request.getScheme(), request.getHeader("host")));
```
pbortnik: i don't think that it is a good idea to put it into rq
you should provide it as a parameter

ihar-kahadouski: fixed

src/main/java/com/epam/ta/reportportal/core/imprt/impl/AbstractImportStrategy.java
```diff
@@ -89,7 +89,7 @@ protected void finishLaunch(String launchId, ReportPortalUser.ProjectDetails pro
 			ParseResults results) {
 		FinishExecutionRQ finishExecutionRQ = new FinishExecutionRQ();
 		finishExecutionRQ.setEndTime(results.getEndTime());
-		finishLaunchHandler.finishLaunch(launchId, finishExecutionRQ, projectDetails, user);
+		finishLaunchHandler.finishLaunch(launchId, finishExecutionRQ, projectDetails, user, null);
```
pbortnik: Check this null

PR #839, opened 2019-07-16
[EPMRPP-39620] Handle request with empty attributes

ihar-kahadouski:


src/main/java/com/epam/ta/reportportal/core/widget/content/loader/CumulativeTrendChartLoader.java
```diff
@@ -80,31 +80,27 @@
 		if (ArrayUtils.isEmpty(attributes)) {
 			content = widgetContentRepository.cumulativeTrendStatistics(filter,
 					contentFields,
-					sort,
-					storedAttributes.get(0),
+					sort, storedAttributes.size() > 0 ? storedAttributes.get(0) : null,
```
avarabyeu: check for null

ihar-kahadouski: checked and fixed

PR #856, opened 2019-07-22
EPMRPP-37503 || Plugin uploading logic update

Yumfriez:


src/main/java/com/epam/ta/reportportal/plugin/ReportPortalExtensionFactory.java
```diff
@@ -0,0 +1,26 @@
+package com.epam.ta.reportportal.plugin;
```
avarabyeu: license headers everywhere

Yumfriez: added

PR #926, opened 2019-08-27
[WIP] Changed existing history method

tatyana222:
Draft work in progress pull request to review intermediate changes


src/main/java/com/epam/ta/reportportal/core/item/impl/history/TestItemsHistoryHandlerImpl.java
```diff
@@ -56,18 +62,35 @@
...
+		Page<Long> testItems = getTestItemHandler.getTestItemsIds(
+				filter,
+				pageable,
+				projectDetails,
+				user,
+				launchId,
+				filterId,
+				launchesLimit
+		);
+		List<Long> itemIds = new ArrayList<>(testItems.getContent());
```
Yumfriez: getContent returns List, so new List creation isn't necessary

tatyana222: It's because getContent returns unmodifiable list. removeAll method does not work on it

PR #957, opened 2019-09-17
EPMRPP-42185 || Initial status attribute impl

Yumfriez:


src/main/java/com/epam/ta/reportportal/core/item/impl/UpdateTestItemHandlerImpl.java
```diff
@@ -270,6 +273,21 @@ public OperationCompletionRS updateTestItem(ReportPortalUser.ProjectDetails proj
 				.collect(toList());
 	}
 
+	private void checkInitialStatusAttribute(TestItem testItem) {
+		Optional<ItemAttribute> statusAttirbute = testItem.getAttributes()
+				.stream()
+				.filter(attribute -> attribute.getKey().equalsIgnoreCase(INITIAL_STATUS_ATTRIBUTE_KEY) && attribute.isSystem())
```
pbortnik: possible npe if attribute key is null
CONSTANT.equals(variable)

src/main/java/com/epam/ta/reportportal/core/item/impl/UpdateTestItemHandlerImpl.java
```diff
@@ -221,6 +223,7 @@ public OperationCompletionRS updateTestItem(ReportPortalUser.ProjectDetails proj
 					INCORRECT_REQUEST,
 					"Unable to change status on test item with children"
 			);
+			checkInitialStatusAttribute(testItem);
```
pbortnik: what about updating some tests according to the new logic?

src/main/java/com/epam/ta/reportportal/core/item/impl/UpdateTestItemHandlerImpl.java
Yumfriez: fixed

src/main/java/com/epam/ta/reportportal/core/item/impl/UpdateTestItemHandlerImpl.java
Yumfriez: added

PR #958, opened 2019-09-17
EPMRPP-41163 || Unique integration name impl

Yumfriez:


src/main/java/com/epam/ta/reportportal/core/integration/impl/CreateIntegrationHandlerImpl.java
```diff
@@ -79,6 +83,11 @@ public EntryCreatedRS createGlobalIntegration(IntegrationRQ createRequest, Strin
 		IntegrationService integrationService = integrationServiceMapping.getOrDefault(integrationType.getName(),
 				this.basicIntegrationService
 		);
+
+		ofNullable(createRequest.getName()).map(String::toLowerCase).ifPresent(name -> {
```
pbortnik: should we do something if name is null ?

src/main/java/com/epam/ta/reportportal/core/integration/impl/CreateIntegrationHandlerImpl.java
```diff
@@ -102,6 +111,11 @@ public EntryCreatedRS createProjectIntegration(String projectName, IntegrationRQ
 				this.basicIntegrationService
 		);
 
+		ofNullable(createRequest.getName()).map(String::toLowerCase).ifPresent(name -> {
```
pbortnik: should we do something if name is null ?

src/main/java/com/epam/ta/reportportal/core/integration/impl/CreateIntegrationHandlerImpl.java
Yumfriez: validation added

src/main/java/com/epam/ta/reportportal/core/integration/impl/CreateIntegrationHandlerImpl.java
Yumfriez: validation added

src/main/java/com/epam/ta/reportportal/core/integration/impl/CreateIntegrationHandlerImpl.java
```diff
@@ -79,6 +84,13 @@ public EntryCreatedRS createGlobalIntegration(IntegrationRQ createRequest, Strin
 		IntegrationService integrationService = integrationServiceMapping.getOrDefault(integrationType.getName(),
 				this.basicIntegrationService
 		);
+
+		String integrationName = ofNullable(createRequest.getName()).map(String::toLowerCase).map(name -> {
```
pbortnik: what about annotations @NotNull and @NotEmpty for fields in IntegrationRQ?

PR #962, opened 2019-09-24
[EPMRPP-42257] Clean logs from analyzer on retry item

ihar-kahadouski:


src/main/java/com/epam/ta/reportportal/core/analyzer/auto/impl/SearchLogServiceImpl.java
```diff
@@ -129,8 +130,10 @@ public SearchLogServiceImpl(ProjectRepository projectRepository, LaunchRepositor
 				value.getLogs().add(TO_LOG_ENTRY.apply(log));
 				return value;
 			});
-			Launch launch = launchRepository.findById(log.getTestItem().getLaunchId())
-					.orElseThrow(() -> new ReportPortalException(ErrorType.LAUNCH_NOT_FOUND, log.getTestItem().getLaunchId()));
+			Long launchId = ofNullable(log.getTestItem()
```
pbortnik: why is it better than the previous variant?

ihar-kahadouski: Retrise doesn't have launch Id and not supported with search log functionality. So we check it here

PR #970, opened 2019-09-25
[EPMRPP-42282] Rerun flag in resource

ihar-kahadouski:


src/main/java/com/epam/ta/reportportal/ws/converter/converters/LaunchConverter.java
```diff
@@ -75,6 +75,7 @@
 		resource.setApproximateDuration(db.getApproximateDuration());
 		resource.setHasRetries(db.isHasRetries());
 		ofNullable(db.getUserId()).flatMap(id -> userRepository.findLoginById(id)).ifPresent(resource::setOwner);
+		resource.setRerun(db.isRerun());
```
pbortnik: please, add to the change set info about the task

PR #973, opened 2019-09-26
Add get user photo endpoint for admin

ihar-kahadouski:


src/main/java/com/epam/ta/reportportal/ws/controller/FileStorageController.java
```diff
@@ -95,6 +97,17 @@ public void getUserPhoto(@PathVariable String projectName, @RequestParam(value =
 		toResponse(response, userPhoto);
 	}
 
+	@Transactional
+	@PreAuthorize(ADMIN_ONLY)
+	@GetMapping("/userphoto")
+	@ApiIgnore
+	public void getUserPhotoByAdmin(@RequestParam(value = "id") String username,
```
pbortnik: don't need a new endpoint. ui just past admin project and we allow him to retrieve any photo

ihar-kahadouski: done

PR #1018, opened 2019-10-24
introduce env variable for queues

pbortnik:


src/main/java/com/epam/ta/reportportal/core/configs/rabbit/ReportingConfiguration.java
```diff
@@ -54,12 +54,8 @@
...
 	public static final String QUEUE_DLQ = "reporting.dlq";
 
-	public static int QUEUE_AMOUNT;
-
-	@Value("${rp.amqp.addresses}")
-	public void setQueueAmount(int queueAmount) {
-		QUEUE_AMOUNT = queueAmount;
-	}
+	@Value("${rp.amqp.queues}")
+	public int QUEUE_AMOUNT;
```
knodwell: let's call varialbe queueAmount as it is not constant any more

PR #1046, opened 2019-11-20
[EPMRPP-44258] Introduce jdk 11

ihar-kahadouski:


sealights.gradle
```diff
@@ -1,14 +1,3 @@
-buildscript {
```
avarabyeu: just to make sure, will sealights plugin be working without that section?

ihar-kahadouski: No

PR #1047, opened 2019-11-20
[EPMRPP-44614] Add validation to launch names input value

ihar-kahadouski:


src/main/java/com/epam/ta/reportportal/core/launch/impl/GetLaunchHandlerImpl.java
```diff
@@ -179,8 +179,8 @@ public LaunchResource getLaunchByProjectName(String projectName, Pageable pageab
 
 	@Override
 	public List<String> getLaunchNames(ReportPortalUser.ProjectDetails projectDetails, String value) {
-		expect(value.length() > 2, equalTo(true)).verify(INCORRECT_FILTER_PARAMETERS,
-				formattedSupplier("Length of the launch name string '{}' is less than 3 symbols", value)
+		expect(value.length() > 0 && value.length() < 257, equalTo(true)).verify(INCORRECT_FILTER_PARAMETERS,
```
pbortnik: what about a constant ? constant should be in commons with others

ihar-kahadouski: done

PR #1055, opened 2019-11-27
[EPMRPP-44730] handle launches with not UUID identificator

ihar-kahadouski:


src/main/java/com/epam/ta/reportportal/util/ReportingQueueService.java
```diff
@@ -40,10 +42,9 @@
 	 * @return
 	 */
 	public String getReportingQueueKey(String launchUuid) {
-		int value = UUID.fromString(launchUuid).hashCode();
+		int value = launchUuid.matches(UUID_REGEX) ? UUID.fromString(launchUuid).hashCode() : launchUuid.hashCode();
```
pbortnik: what about updating existing unit test or creating one if not exists?

ihar-kahadouski: done

PR #1066, opened 2019-12-05
Integration name added to the integration-related activities

Yumfriez:


src/main/java/com/epam/ta/reportportal/core/events/activity/IntegrationCreatedEvent.java
```diff
@@ -56,6 +65,7 @@ public Activity toActivity() {
 				.addObjectId(integrationActivityResource.getId())
 				.addObjectName(integrationActivityResource.getTypeName())
 				.addProjectId(integrationActivityResource.getProjectId())
+				.addHistoryField(Optional.of(integrationNameField))
```
pbortnik: can be used 'addHistoryField(String field, String before, String after)' here, so we can remove upper above

src/main/java/com/epam/ta/reportportal/core/events/activity/IntegrationDeletedEvent.java
```diff
@@ -56,6 +65,7 @@ public Activity toActivity() {
 				.addObjectId(integrationActivityResource.getId())
 				.addObjectName(integrationActivityResource.getTypeName())
 				.addProjectId(integrationActivityResource.getProjectId())
+				.addHistoryField(Optional.of(integrationNameField))
```
pbortnik: can be used 'addHistoryField(String field, String before, String after)' here, so we can remove upper above

src/main/java/com/epam/ta/reportportal/core/events/activity/IntegrationCreatedEvent.java
Yumfriez: fixed

src/main/java/com/epam/ta/reportportal/core/events/activity/IntegrationDeletedEvent.java
Yumfriez: fixed

PR #1069, opened 2019-12-10
fix npe

maartenjanvangool:
Get a bit more information when rabbitmq cannot be found instead of a NPE.


maartenjanvangool:
Thanks!

PR #1077, opened 2019-12-13
[EPMRPP-45048] - Async email sending

pbortnik:


src/main/resources/application.yaml
```diff
@@ -86,7 +86,7 @@ rp:
     host: postgres
     port: 5432
     user: rpuser
-    pass: rppass
+    pass: rppass123
```
Yumfriez: probably should be changed back

src/main/java/com/epam/ta/reportportal/core/configs/EmailConfiguration.java
```diff
@@ -37,6 +40,11 @@
 	//		return new MailServiceFactory(templateEngine, encryptor, settingsRepository);
 	//	}
 
+	@Bean
+	public ExecutorService emailExecutorService() {
+		return Executors.newSingleThreadExecutor();
```
Yumfriez: can be changed to Spring's TaskExecutor (thread-pooling) so in future we can use it for different notifications

PR #1079, opened 2019-12-17
[EPMRPP-45078] Fix npe on finish item without launch in async controller

ihar-kahadouski:


src/main/java/com/epam/ta/reportportal/ws/controller/TestItemAsyncController.java
```diff
@@ -87,6 +86,10 @@ public EntryCreatedAsyncRS startChildItem(@PathVariable String projectName, @Aut
 	@PreAuthorize(ALLOWED_TO_REPORT)
 	public OperationCompletionRS finishTestItem(@PathVariable String projectName, @AuthenticationPrincipal ReportPortalUser user,
 			@PathVariable String testItemId, @RequestBody @Validated FinishTestItemRQ finishExecutionRQ) {
+		expect(finishExecutionRQ.getLaunchUuid(), StringUtils::isNotEmpty).verify(
```
pbortnik: can we move it somewhere into handler?

ihar-kahadouski: yes, done

PR #1085, opened 2019-12-23
Default 'NULL' value provided for parameter resource

Yumfriez:


src/main/java/com/epam/ta/reportportal/ws/converter/builders/TestItemBuilder.java
```diff
@@ -152,7 +154,7 @@ public TestItemBuilder addParameters(List<ParameterResource> parameters) {
 			testItem.setParameters(parameters.stream().map(it -> {
 				Parameter parameter = new Parameter();
 				parameter.setKey(it.getKey());
-				parameter.setValue(it.getValue());
+				parameter.setValue(ofNullable(it.getValue()).orElse(PARAMETER_NULL_VALUE));
```
pbortnik: i think that it's not expected behaviour, so it should be validated somewhere on the request level

PR #1087, opened 2019-12-24
[EPMRPP-45182] - Queues pod upgrade fix

pbortnik:


src/main/java/com/epam/ta/reportportal/core/configs/rabbit/ReportingConfiguration.java
```diff
@@ -178,38 +177,25 @@ public Queue queueDlq(AmqpAdmin amqpAdmin) {
...
-			}
-		}
-		if (containers.size() < queuesPerPod) {
-			logger.error("Started amount of consumers less then configured (or single node start) : {} < {}", containers.size(), queuesPerPod);
+		int consumersCount = 0;
+		while (consumersCount < queuesPerPod) {
+			SimpleMessageListenerContainer listenerContainer = new SimpleMessageListenerContainer(connectionFactory);
+			containers.add(listenerContainer);
+			listenerContainer.setConnectionFactory(connectionFactory);
+			listenerContainer.addQueues(queues.get(consumersCount));
```
knodwell: don't know for sure, but I have read that it's needed to use addQueueNames() instead of addQueues() here, otherwise some initialization problems possible

PR #1088, opened 2019-12-24
Support v4 get rq by string id

pbortnik:
fixes [#782](https://github.com/reportportal/reportportal/issues/782)


Budda0ne (commented):
...

PR #1093, opened 2019-12-27
fix xls reports

ihar-kahadouski:


build.gradle
```diff
@@ -92,6 +92,9 @@ dependencies {
 
     compile 'com.sun.mail:javax.mail:1.6.2'
     compile 'net.sf.jasperreports:jasperreports:6.10.0'
+//    JasperReport's export to XLS uses Apache POI. Reports 6.3.1 needs version NOT
+//    higher than 3.14 since 3.15 contains breaking API changes
```
pbortnik: do we still need these comments ?

PR #1128, opened 2020-02-06
EPMRPP-45508 [WS] Secret JWT token

pbortnik:


src/main/java/com/epam/ta/reportportal/core/configs/SecurityConfiguration.java
```diff
@@ -86,15 +95,20 @@ protected MethodSecurityExpressionHandler createExpressionHandler() {
 	@EnableResourceServer
 	public static class SecurityServerConfiguration extends ResourceServerConfigurerAdapter {
 
+		private static final String SECRET_KEY = "secret.key";
+
+		@Autowired
```
Yumfriez: `@Autowired` is for beans only

pbortnik: done

PR #1136, opened 2020-02-13
[EPMRPP-46274] Change encryption type

ihar-kahadouski:


src/main/java/com/epam/ta/reportportal/core/user/UserPasswordService.java
```diff
@@ -0,0 +1,34 @@
...
+
+import com.epam.ta.reportportal.entity.user.User;
+import com.epam.ta.reportportal.ws.model.user.ChangePasswordRQ;
+import com.epam.ta.reportportal.ws.model.user.CreateUserRQFull;
+import com.epam.ta.reportportal.ws.model.user.ResetPasswordRQ;
+
+/**
+ * @author <a href="mailto:ihar_kahadouski@epam.com">Ihar Kahadouski</a>
+ */
+public interface UserPasswordService {
```
pbortnik: Can't see need of this interfacel. The idea of interfaces that it can be used in several places. But you created methods that does the same (password encoding) but gets different parameters. If you want to create an interface for it, so it's better to create one with the method 'encode', that receives string as parameter and returns encoded string back.

ihar-kahadouski: fixed

PR #1139, opened 2020-02-16
EPMRPP-46188 || Nested steps status hierarchy calculation refactoring

Yumfriez:


src/main/java/com/epam/ta/reportportal/core/analyzer/auto/impl/SearchLogServiceImpl.java
```diff
@@ -108,6 +108,7 @@ public SearchLogServiceImpl(ProjectRepository projectRepository, LaunchRepositor
 
 		searchRq.setFilteredLaunchIds(searchCollectorFactory.getCollector(searchMode).collect(request.getFilterId(), launch));
 
+		//TODO fix query
```
pbortnik: describe what is the problem with this query and how to fix it

Yumfriez: updated

PR #1157, opened 2020-03-02
Update Dockerfile-release

tomassatka:
Motivation for the change is to run reportportal on openshift platform. Vanilla openshift setup doesnt allow to run containers as root and will generate random users from given range. Therefore before running reportportal on our side we have to modify the images to remove the requirement of root user to run the container. 

Also not sure if we have to add one more command to mkdir /data and then adding writing permissions 

One more thing.. Openshift doesnt support kubernetes ingress route modifications that are setup as:
#    kubernetes.io/ingress.class: nginx
#    nginx.ingress.kubernetes.io/ssl-redirect: "false"
#    nginx.ingress.kubernetes.io/rewrite-target: /$2
#    nginx.ingress.kubernetes.io/x-forwarded-prefix: /$1

Therefore we had to start independent nginx inside namespace where we inject nginx.conf over configMap to provide following configuration (including snippet):
       
   location / {
                proxy_pass http://reportportal-index:8080;
            }

            location /ui/ {
                proxy_pass http://reportportal-ui:8080;
                rewrite /ui/(.*) /$1 break;
                proxy_redirect off;
                proxy_set_header Host $host;
            }

            location /api/ {
                proxy_pass http://reportportal-api:8585;
                rewrite /api/(.*) /$1 break;
                proxy_redirect off;
                proxy_set_header Host $host;
            }

            location /uat/ {
                proxy_pass http://reportportal-uat:9999;
                rewrite /uat/(.*) /$1 break;
                proxy_redirect off;
                proxy_set_header Host $host;
            }

Could you please also comment on the approach? If that can be integrated as an option or setup traefik within Openshift..


docker/Dockerfile-release
```diff
@@ -15,7 +15,7 @@ exec java \$JAVA_OPTS -jar \$APP_FILE' > /start.sh && chmod +x /start.sh"
 
 VOLUME /tmp
 
-RUN wget -O /$APP_FILE $APP_DOWNLOAD_URL
+RUN wget -O /$APP_FILE $APP_DOWNLOAD_URL && mkdir /plugins && chmod a+w /var /plugins
```
HardNorth: Two questions:
1. You create and apply permissions on `plugins` folder because plugins uploads in this folder?
2. You apply permissions on `/var` due to file logging?

tomassatka: 1/ correct
2/ correct

also might be that /data folder is required and needs permissions

tomassatka: basically good test is to add step into dockerfile

USER nobody 

and try to start

tomassatka: 1/ correct
2/ correct

Also directory /data might be required with permissions on it

Basically good test is to add into docker file line

USER nobody 

and try to start the application properly

HardNorth: 1. I believe we already moved all plugin uploads to Minio (probably in current development version, 5.1), and there shouldn't be any other option to upload plugins straight on disc in cluster deployments. Cloud applications shouldn't consider itself as a storage.
2. Then we need to disable any log output, the same as for p.1. I created internal ticket for this.

The same for `/data` directory. It should be mounted as an external volume with permissions applied on Kubernetes deployment chart level.

tomassatka: One option would be to add persistanceVolumeClaim as template:

apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: api-service-data
spec:
  storageClassName: nfs
  accessModes:
    - ReadWriteOnce
  resources:
    requests:
      storage: {{ $.Values.reportportal.data.persistentVolumeClaimSize }}

And then mount it to locations
 volumeMounts:
            - mountPath: /data
              name: api-service-data

and define volumes as:
volumes:
        - name: api-service-data
          persistentVolumeClaim:
            claimName: api-service-data

Could work like that..

HardNorth: We removed any logging into files by default. And your suggestion for plugins is for kubernetes deployment: https://github.com/reportportal/kubernetes

tomassatka: is that now integrated for next release 5.1?
also this https://github.com/reportportal/migrations/pull/140 ?

thanks

waynesun09: No need update for /var dir with latest 5.1 as tested.
For /plugins dir even though data will be stored via minio, still need the group write access to the dir, may update with 'g+w' rather than 'a+w'.

PR #1179, opened 2020-03-27
Fix start script in Dockerfiles with openjdk:11-jre-slim base image

ivan-sukhomlyn:
@HardNorth I've got the mistake with startup script definition in Dockerfiles with `openjdk:11-jre-slim`
It's because I've used Dockerfile in the root folder for testing and it contains `alpine` image as basic.
So, that's why we have different `echo` command outputs
Sorry for leading you to wrong way


ivan-sukhomlyn:
@HardNorth Could you please take a look at the PR?

HardNorth:
Well, already did that: https://github.com/reportportal/service-api/commit/e41e6d17892993ad2625013268fcb4d3c5a99fc9

PR #1208, opened 2020-05-11
Add group write permission to default /plugins dir

waynesun09:
Even though the data will be stored via minio, still the app need
permission to add the contents under the dir when run in rootless
container with group access env like OpenShift.
Signed-off-by: Wayne Sun <gsun@redhat.com>


waynesun09:
Close as duplicate with:
https://github.com/reportportal/service-api/pull/1157

PR #1217, opened 2020-05-20
Expose prometheus endpoints without authentication

waynesun09:
PR#1053 have switched to Spring prometheus metrics, and enabled
the endpoints /metrics and /prometheus.

For prometheus scraping it should not require authentication, so
expose the endpoints like /health.

Signed-off-by: Wayne Sun <gsun@redhat.com>


HardNorth:
@waynesun09 Prometheus has  ['bearer_token'](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#scrape_config)  configuration. And you can use admin token for metric scrape. Your patch opens metrics to the whole world, which is not a good idea due to security reasons.

So no, I'm not going to approve this in any case.

PR #1319, opened 2020-11-02
Fix launch locking performance regression

coufalja:
Fixes https://github.com/reportportal/reportportal/issues/950
Fixes https://github.com/reportportal/reportportal/issues/1186


coufalja:
@Yumfriez @pbortnik Would you mind taking a look? (and to the linked issue as well)

pbortnik:
Hi @coufalja, 
Those "selectForUpdate" was implemented intentionally. We can't simply remove it. It will break our reporting pipeline. Now, we are working on refactoring of the logic to avoid locks with a different approach, so the fix is going to be released in nearest releases.

PR #1352, opened 2021-01-18
[EPMRPP-58502] - Save a path for attachments w/o logs, and than find them in Minio and delete in Minio

pbortnik:


src/test/java/com/epam/ta/reportportal/ws/rabbit/AttachmentConsumerTest.java
```diff
@@ -64,4 +64,16 @@ void consume() {
...
+	@Test
+	void consumePaths() {
+		ArrayList<String> paths = Lists.newArrayList("fileId");
+		DeleteAttachmentEvent event = new DeleteAttachmentEvent();
+		event.setPaths(paths);
+
+		attachmentConsumer.onEvent(event);
+
+		verify(dataStoreService, times(1)).delete(any());
+		verify(attachmentRepository, times(1)).deleteAllByIds(any());
```
miracle8484, line 77: Better to use more specific check, i mean arguments from **delete(any())** and **deleteAllByIds(any())** change to specific values instead of **any()**
Could you change it, please.

src/main/java/com/epam/ta/reportportal/core/file/DeleteFilesHandler.java
```diff
@@ -0,0 +1,75 @@
...
+
+	@Autowired
+	private MessageBus messageBus;
+
+	public OperationCompletionRS removeFilesByCsv(MultipartFile file) {
+		CSVParser parser = new CSVParserBuilder().withSeparator(',').withIgnoreQuotations(true).build();
+		try (CSVReader csvReader = new CSVReaderBuilder(new InputStreamReader(file.getInputStream(), StandardCharsets.UTF_8)).withSkipLines(
+				CSV_SKIP_LINES).withCSVParser(parser).build()) {
+			List<String[]> attachments = csvReader.readAll();
+			List<String> pathsForDelete = Lists.newArrayListWithCapacity(attachments.size());
```
miracle8484, line 57: Is there possible NPE or **csvReader.readAll()** mustn't return null in any case?

pbortnik: It returns empty list in case empty csv file.

src/test/java/com/epam/ta/reportportal/ws/rabbit/AttachmentConsumerTest.java
pbortnik: done

src/main/java/com/epam/ta/reportportal/core/file/DeleteFilesHandler.java
miracle8484: ok, thx, but question not only for empty csv file, i about any case, if won't be null - than ok, if not, better to add check.

src/test/java/com/epam/ta/reportportal/ws/rabbit/AttachmentConsumerTest.java
miracle8484: ok, thx

src/main/java/com/epam/ta/reportportal/core/file/DeleteFilesHandler.java
miracle8484: Disscussed, ok

miracle8484 (approved):

PR #1370, opened 2021-04-07
[EPMRPP-60810] Refactoring of manual deletion of Projects/ Launches/ Item( Suites/Tests/Steps)

pbortnik:


src/main/java/com/epam/ta/reportportal/core/launch/impl/DeleteLaunchHandlerImpl.java
```diff
@@ -121,7 +121,7 @@ public DeleteBulkRS deleteLaunches(DeleteBulkRQ deleteBulkRQ, ReportPortalUser.P
 
 		launchRepository.deleteAll(toDelete);
 		toDelete.stream().map(TO_ACTIVITY_RESOURCE).forEach(it -> {
-			eventPublisher.publishEvent(new DeleteLaunchAttachmentsEvent(it.getId()));
+			attachmentRepository.moveForDeletionByLaunchId(it.getId());
 			messageBus.publishActivity(new LaunchDeletedEvent(it, user.getUserId(), user.getUsername()));
```
miracle8484, line 125: Do we really need this sent this event?

src/main/java/com/epam/ta/reportportal/core/log/impl/DeleteLogHandlerImpl.java
```diff
@@ -85,7 +84,8 @@ public OperationCompletionRS deleteLog(Long logId, ReportPortalUser.ProjectDetai
 		Log log = validate(logId, user, projectDetails);
 		try {
 			logRepository.delete(log);
-			cleanUpLogData(log);
+			ofNullable(log.getAttachment()).ifPresent(attachment -> attachmentRepository.moveForDeletion(Collections.singletonList(
+					attachment.getId())));
```
miracle8484, line 88: May be little update and add appropriate method to repository will be better
if (log.getAttachment())  {
    attachmentRepository.moveForDeletion(log.getAttachment().getId());
}

pbortnik: Method replaced with more appropriate

src/main/java/com/epam/ta/reportportal/core/launch/impl/DeleteLaunchHandlerImpl.java
pbortnik: Event doesn't relate to deletion. It is logic of storing reportportal activities.
Method is replaced with more appropriate.

miracle8484: It depends of the point of view, i thinks if we implement new logic, it should be clean for us - it's deletions from database + moving data to "deletion table". But if we have some extensible point we may have some unpredictable behavior on it. And for now, when we are trying to simplify this flow and improve performance (it's important, we try to simplify, not to make more extensible and so on). Will be better if we remove all unused events(if it's unused) too.

miracle8484: Discussed

miracle8484 (approved):

PR #1400, opened 2021-06-21
EPMRPP-62926  Change manual deletion logic to delete logs from analyzer by launch_id

pbortnik:


src/main/java/com/epam/ta/reportportal/core/launch/impl/DeleteLaunchHandlerImpl.java
```diff
@@ -82,12 +81,9 @@ public OperationCompletionRS deleteLaunch(Long launchId, ReportPortalUser.Projec
 		Launch launch = launchRepository.findById(launchId)
 				.orElseThrow(() -> new ReportPortalException(ErrorType.LAUNCH_NOT_FOUND, launchId));
 		validate(launch, user, projectDetails);
-		logIndexer.cleanIndex(projectDetails.getProjectId(),
-				logRepository.findItemLogIdsByLaunchIdAndLogLevelGte(launchId, LogLevel.ERROR.toInt())
-		);
 
+		logIndexer.indexLaunchesRemove(projectDetails.getProjectId(), Lists.newArrayList(launchId));
```
miracle8484, line 85: NIT: may be will be better to make appropriate methods(for logIndexer), for instance with single Long launchId, and if it is not possible to send different objects to queue(with list<Long> and simple log), that be normally processed from the analyzer side, than may be resolve this situation into IndexLaunchRemove (with another constructor) will be better.

miracle8484 (approved):

PR #1521, opened 2021-12-15
EPMRPP-72478 || Log4j dependency updated to 2.16.0

Yumfriez:


Yumfriez:
> Another vote for this one - Log4j 2.15 does not fully mitigate the original exploit.

Yep, so we updated to 2.16.0

PR #1596, opened 2022-06-23
EPMRPP-77972 ||  Add missed nested logs in recalculation on deletion

pbortnik:


src/main/java/com/epam/ta/reportportal/core/ElementsCounterService.java
```diff
@@ -31,9 +32,11 @@
 @Service
 public class ElementsCounterService {
 
-	private TestItemRepository testItemRepository;
+	private static final int BATCH_SIZE = 50;
```
miracle8484, line 35: Could you move it to variables please, could be needed for more convenient for perf testing

src/main/java/com/epam/ta/reportportal/core/ElementsCounterService.java
```diff
@@ -42,33 +45,41 @@ public ElementsCounterService(TestItemRepository testItemRepository, LogReposito
...
+		final AtomicLong resultedNumber = new AtomicLong(1L);
 		final List<Long> testItemIdsByLaunchId = testItemRepository.findIdsByLaunchId(launchId);
-		resultedNumber += testItemIdsByLaunchId.size();
-		resultedNumber += logRepository.countLogsByTestItemItemIdIn(testItemIdsByLaunchId);
-		resultedNumber += logRepository.countLogsByLaunchId(launchId);
-		return resultedNumber;
+		resultedNumber.addAndGet(testItemIdsByLaunchId.size());
+		Lists.partition(testItemIdsByLaunchId, BATCH_SIZE).forEach(batch -> {
+			resultedNumber.addAndGet(logRepository.countLogsByTestItemItemIdIn(testItemIdsByLaunchId));
+			resultedNumber.addAndGet(logRepository.countLogsByLaunchId(launchId));
```
miracle8484, line 53: May be we should add logs for launchId only one time?

pbortnik: There are two different types of logs.
Launch logs that applied to launch object
And test item logs that applied to test items inside that launch
So, we need to count both of them

src/main/java/com/epam/ta/reportportal/core/ElementsCounterService.java
pbortnik: done

miracle8484 (approved):

PR #1627, opened 2022-10-13
EPMRPP-79768 || Update creating project

APiankouski:


APiankouski:
Related pr for commons-model: https://github.com/reportportal/commons-model/pull/324

PR #1638, opened 2022-11-25
EPMRPP-78657 || Notification of an inexistent launch in Edit/Duplicate

APiankouski:


APiankouski:
Allow v1/{projectName}/launch/names return all launch names without filters

src/main/java/com/epam/ta/reportportal/core/launch/impl/GetLaunchHandlerImpl.java
```diff
@@ -226,7 +226,7 @@ private Iterable<LaunchResource> getLaunchResources(Page<Launch> launches) {
 
 	@Override
 	public List<String> getLaunchNames(ReportPortalUser.ProjectDetails projectDetails, String value) {
-		expect(value.length() >= MIN_LAUNCH_NAME_LENGTH && value.length() <= MAX_LAUNCH_NAME_LENGTH, equalTo(true)).verify(
```
IvanKustau, line 229: Why we removing MIN_LAUNCH_NAME_LENGTH here? And if we need to remove it, then you also should remove this from log message

APiankouski: Will be fixed, thx

PR #1645, opened 2022-12-20
EPMRPP-81019 || Launch import. Error 'org.hibernate.exception.DataException: could not execute statement' occurs on importing launch

APiankouski:


src/main/java/com/epam/ta/reportportal/core/imprt/impl/junit/XunitImportHandler.java
```diff
@@ -220,9 +223,9 @@ private void startTestItem(String name) {
 	private void startStepItem(String name, String startTime, String duration) {
 		StartTestItemRQ rq = new StartTestItemRQ();
 		rq.setLaunchUuid(launchUuid);
-		rq.setStartTime(EntityUtils.TO_DATE.apply(parseTimeStamp(startTime)));
+		rq.setStartTime(EntityUtils.TO_DATE.apply(startTime != null ? parseTimeStamp(startTime) : startItemTime));
```
miracle8484, line 226: .setStartTime(EntityUtils.TO_DATE.apply(startTime != null ? parseTimeStamp(startTime) : null));

miracle8484 (approved):

PR #1737, opened 2023-07-24
EPMRPP-82848 || Add new types of event on user role change

rkukharenka:


src/main/java/com/epam/ta/reportportal/core/project/impl/UpdateProjectHandlerImpl.java
```diff
@@ -476,11 +477,25 @@ private void updateProjectUserRoles(Map<String, String> userRoles, Project proje
                 ErrorType.ACCESS_DENIED);
           }
         }
+        String oldRole = updatingProjectUser.get().getProjectRole().getRoleName();
         updatingProjectUser.get().setProjectRole(newProjectRole.get());
+        String newRole = updatingProjectUser.get().getProjectRole().getRoleName();
+
+        ChangeRoleEvent changeRoleEvent = getChangeRoleEvent(updatingProjectUser.get().getUser(),
```
APiankouski, line 484: eventPublishing can be split to separate methods.

PR #1818, opened 2023-10-09
EPMRPP-86219 || Implement immediate AutoAnalyze on TestItemFinishing

APiankouski:


src/main/java/com/epam/ta/reportportal/core/events/handler/item/TestItemAutoAnalysisRunner.java
```diff
@@ -0,0 +1,81 @@
...
+  private void logIndex(TestItem testItem, Long projectId, AnalyzerConfig config) {
+    logIndexer.indexItemsLogs(projectId, testItem.getLaunchId(), List.of(testItem.getItemId()),
+        config);
+  }
+
+  private boolean isContainAttribute(TestItem testItem) {
+    Optional<ItemAttribute> first = testItem.getAttributes().stream()
+        .filter(it -> IMMEDIATE_AUTO_ANALYSIS.equals(it.getKey())).findFirst();
+    return first.isPresent() && Boolean.parseBoolean(first.get().getValue());
+  }
```
IvanKustau, line 80: You can rewrite it to use anyMatch like this:
```
    return testItem.getAttributes().stream()
        .anyMatch(it -> IMMEDIATE_AUTO_ANALYSIS.equals(it.getKey()) && Boolean.parseBoolean(it.getValue()));
```

Also the method name should be changed to something like containsImmediateAutoAnalysisAttribute

APiankouski: Fixed

IvanKustau (approved):

PR #1841, opened 2023-10-25
Use default HttpMessageConverter

janmekota:
Remove configureMessageConverters to use default HttpMessageConverter that properly handles

- application/json
- text/plain
- application/openmetrics-text

https://github.com/reportportal/reportportal/issues/1999


IvanKustau:
@janmekota We can't merge this pull request, because these changes will break backwards compatibility in our project and can cause different issues. Instead we are going to add new message converter for your case. You can check it here:
https://github.com/reportportal/service-api/pull/1886

Kozovskaya:
Hello @janmekota 
I'm pleased to let you know that the fix is already released!
https://github.com/reportportal/reportportal/releases/tag/v24.1

PR #1854, opened 2023-11-09
Plugin System refactoring

Yumfriez:
Refactoring of Plugin System.

I redesigned PluginManager to a lightweight container with simple logic.

All functionality (validation, metadata resolving, file uploads) was moved to different services.

@pbortnik 🐰


src/main/java/com/epam/ta/reportportal/core/integration/bootstrap/DefaultPluginBootstrapper.java
```diff
@@ -0,0 +1,50 @@
...
+		this.pluginLoader = pluginLoader;
+	}
+
+	@Override
+	@PostConstruct
+	public void startUp() {
+		// load and start all enabled plugins of application
+		integrationTypeRepository.findAll()
+				.stream()
+				.filter(IntegrationType::isEnabled)
```
pbortnik, line 39: Probably we must start disabled plugins as well. In future we can enable it, but we won't have loaded plugin?

pbortnik (changes requested):

src/main/java/com/epam/ta/reportportal/core/integration/bootstrap/DefaultPluginBootstrapper.java
Yumfriez: LoadPluginsJob will load plugins that were enabled later

pbortnik (approved):

PR #1928, opened 2024-02-23
EPMRPP-89396 || EPMRPP-89397 || EPMRPP-89395 || Change batch endpoint

APiankouski:


src/main/java/com/epam/ta/reportportal/core/launch/DeleteLaunchHandler.java
```diff
@@ -51,4 +52,15 @@ OperationCompletionRS deleteLaunch(Long launchId, ReportPortalUser.ProjectDetail
...
+
+  /**
+   * Bulk launches delete.
+   *
+   * @param ids   list of launch ids
+   * @param projectDetails Project Details
+   * @param user           User
+   * @return DeleteLaunchesRS
+   */
+  DeleteBulkRS deleteLaunches(List<Long> ids,
```
grabsefx, line 64: interface implementation class should be updated accordingly

APiankouski: DeleteBulkRS - response of api don't change

PR #1932, opened 2024-02-28
EPMRPP-80927 || Improve merge algorithm

APiankouski:


src/main/java/com/epam/ta/reportportal/core/item/identity/TestItemUniqueIdGenerator.java
```diff
@@ -84,10 +109,31 @@ private String prepareForEncoding(TestItem testItem, List<Long> parentIds, Launc
...
+      System.out.println("TestItem id " + id);
+      if (testItem == null) {
+        dbQueries++;
+        System.out.println("TestItemUniqueIdGenerator db queries: " + dbQueries);
+        testItem = testItemRepository.getById(id);
+        itemsCache.put(id, testItem);
+        CacheStats stats = itemsCache.stats();
+        System.out.println("Cache status: " + stats.toString());
+      } else {
+        System.out.println("Founded in cache");
```
grabsefx, line 133: - please use logger instead
- "Found in cache" might be correct message

src/main/java/com/epam/ta/reportportal/core/item/identity/TestItemUniqueIdGenerator.java
```diff
@@ -84,10 +109,31 @@ private String prepareForEncoding(TestItem testItem, List<Long> parentIds, Launc
...
+
+  private List<TestItem> getTestItems(List<Long> parentIds) {
+    List<TestItem> testItems = new ArrayList<>();
+    for(Long id: parentIds) {
+      TestItem testItem = itemsCache.getIfPresent(id);
+      System.out.println("TestItem id " + id);
+      if (testItem == null) {
+        dbQueries++;
+        System.out.println("TestItemUniqueIdGenerator db queries: " + dbQueries);
+        testItem = testItemRepository.getById(id);
```
grabsefx, line 128: Usually such calling the database in a loop is a place for performance improvement. Perhaps these calls could be executed in paralell or by single request. Also `parentIds` variable could be of `Set<>` type in order to avoid duplicates

PR #1944, opened 2024-03-13
EPMRPP-89659 || Update Notification tab

APiankouski:


Dockerfile
```diff
@@ -10,7 +10,7 @@ RUN if [ "${RELEASE_MODE}" = true ]; then \
     else gradle build --exclude-task test -Dorg.gradle.project.version=${APP_VERSION}; fi
 
 # For ARM build use flag: `--platform linux/arm64`
-FROM --platform=$BUILDPLATFORM amazoncorretto:21.0.2
```
IvanKustau, line 13: Why do we remove Buildplatform from Dockerfile?

pbortnik: up

APiankouski: Alredy fixed

pbortnik (approved):

PR #2264, opened 2025-05-22
EPMRPP-99095 || implemented patch project operation

grabsefx:


src/main/java/com/epam/ta/reportportal/core/project/patch/PatchProjectHandler.java
```diff
@@ -0,0 +1,71 @@
...
+import java.util.List;
+import lombok.extern.log4j.Log4j2;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Service;
+
+@Service
+@Log4j2
+public class PatchProjectHandler {
+
+  private final PatchProjectNameHandler patchProjectName;
```
APiankouski, line 34: Name like it is just field)
probably better reanme to patchProjectNameHandler

PR #2486, opened 2025-12-03
EPMRPP-109652 || Events refactoring

Evelina02:


grabsefx (dismissed):

src/main/java/com/epam/reportportal/ws/rabbit/ProjectLogTypesInitializer.java
```diff
@@ -44,33 +44,34 @@
 public class ProjectLogTypesInitializer {
```
pbortnik: Let’s give this class a more generic name. I think different functionality could be applied after the project creation event, not only log types init.

pbortnik (changes requested):

src/main/java/com/epam/reportportal/ws/rabbit/ProjectLogTypesInitializer.java
Evelina02: renamed to ProjectCreatedConsumer

grabsefx (approved):

PR #2661, opened 2026-04-29
EPMRPP-114585 || restore custom validators

grabsefx:


src/main/java/com/epam/reportportal/base/model/user/EditUserRQ.java
```diff
@@ -39,7 +41,7 @@
 @JsonInclude(Include.NON_NULL)
 public class EditUserRQ {
 
-  @NotBlank
+  @NotBlankString
   @JsonProperty(value = "email")
   private String email;
```
Evelina02, line 46: do we really need a custom annotation here? sounds very similar to existing NotBlank but allows nulls, a bit confusing 🤔

pbortnik (approved):

src/main/java/com/epam/reportportal/base/model/user/EditUserRQ.java
grabsefx: could be switched to the regex pattern which allows null values. I have reverted this model to previous state just to avoid more side effects

Evelina02 (approved):

PR #2761, opened 2026-07-24
Fix: Ambiguous match found for optional joined fields in TMS mapper

imashkanov:
### Description
Fixes bug EPMRPP-117920 where `TMS_TEST_CASE_EXECUTION_MAPPER` throws an `IllegalArgumentException: Ambiguous match found` when mapping a database record to the `TmsTestCaseExecution` entity.

### Root Cause
In jOOQ, when calling `r.field(TableField)` or `r.get(TableField)`, jOOQ first tries to find an exact match. If the table is not joined (meaning the exact field is not in the record), jOOQ falls back to matching by the unqualified column name (e.g., `"id"`). If it finds multiple columns with that name in the current record (like `"tms_test_case_execution"."id"` and `"ticket"."id"`), it throws an ambiguous match exception.

### Fix
Introduced `hasField` and `getFieldValue` utility methods in `RecordMapperUtils` that iterate over the record's fields and check for an exact match using `.equals()`. This bypasses jOOQ's dangerous fallback mechanism. Updated `TMS_TEST_CASE_EXECUTION_MAPPER` to use these safe lookup methods for optional joined fields.

### DIFF
--- a/src/main/java/com/epam/reportportal/base/infrastructure/persistence/dao/util/RecordMapperUtils.java
+++ b/src/main/java/com/epam/reportportal/base/infrastructure/persistence/dao/util/RecordMapperUtils.java
@@ -19,6 +19,7 @@ package com.epam.reportportal.base.infrastructure.persistence.dao.util;
 import java.util.Arrays;
 import java.util.function.Predicate;
 import org.jooq.Field;
+import org.jooq.Record;
 
 /**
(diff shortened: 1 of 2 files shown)


grabsefx (approved):

imashkanov:
already merged https://github.com/reportportal/service-api/commit/0f29b144f63ea501ca29595cac8d0d3fe0744910

PR #2786, opened 2026-08-16
Fix EPMRPP-118571: Return unique lowercase tags in suggestions

imashkanov:
## Summary by CodeRabbit

* **Bug Fixes**
  * Attribute keys are now returned in a consistent lowercase format when retrieved through the TMS repository.


imashkanov:
The changes in this PR correctly address the issue described in EPMRPP-118571. By using `LOWER(key)` and `LOWER(value)` in the distinct queries, the autocomplete endpoints will return a unique list of lowercase tags, preventing duplicate suggestions for tags with different cases.

The implementation looks good and aligns with the expected behavior.

pbortnik (approved):

PR #2787, opened 2026-08-16
Fix/epmrpp 118642 remove status from manual launches

imashkanov:


imashkanov:
I have updated the PR with the requested changes:
1. Removed `@NotNull` and `required = true` from the `status` field in `LaunchResource.java`.
2. Updated the `TO_RESOURCE` function in `LaunchConverter.java` to set `status` to `null` if `db.getLaunchType() == LaunchTypeEnum.MANUAL`.

This ensures that the `status` field is omitted from REST / MCP / WS responses for Manual Launches, while remaining unchanged for AUTO and other non-manual types.

imashkanov:
I have updated the PR to fix the SonarQube issue. The nested ternary operation in `LaunchConverter.java` has been extracted into an independent `if-else` statement.

pbortnik (approved):

PR #2827, opened 2026-09-15
[EPMRPP-121531] Normalize tags imported from CSV

imashkanov:
## Summary
- normalize CSV-imported TMS tag keys using `Locale.ROOT` before deduplication
- add regression coverage for mixed-case, duplicate CSV labels

## Validation
- Added unit test for `Smoke;REGRESSION;smoke` -> `smoke`, `regression`.

This aligns persisted keys for newly imported test cases with the lowercase tag values used by TMS filtering.

## Summary by CodeRabbit

- **Bug Fixes**
  - CSV-imported test case labels are now normalized to lowercase, ensuring consistent label names regardless of capitalization.
  - Duplicate labels that differ only by capitalization are consolidated.

- **Tests**
  - Added coverage confirming case-insensitive label normalization and deduplication during CSV import.


Evelina02 (approved):

src/main/java/com/epam/reportportal/base/core/tms/mapper/importer/TmsTestCaseCsvImporter.java
grabsefx, line 164: could hardcoded path separator cause the OS dependent issue as well?

imashkanov: should not, / is agreed with product folder path separator for csv files for the tms test cases

grabsefx (approved):

